PR #4, opened 2018-12-02
Trying to get C&B to compile with only api.

marchermans:
Moves the compile time dependency to API only.
And adds a runtime dependency to the full mod for testing.


Raycoms:
:shipit:

PR #8, opened 2019-01-07
Render ticks

someaddons:
Removed unneeded cache, now we're storing items based on instances instead. (Currently 2 vs the 50 of the cache). Added the buffer deletion to the removal.

Review pls


marchermans (changes requested):
What makes you think that after reverting the PR that screwed up the cache in the first place, now it is good idea to just remove them?
The cache is their in the first place to clean up things that we do not need, and secondly to save on memory.

What is the goddamn reason why you want an instance render handler?

PR #9, opened 2019-01-09
Feature/optifine

marchermans:
This enables the use of shaders when using Structurize together with Optifine.

Most features are supported, but certain things will produce visual artifacts, like hard shadows on the sides turned away from the sun. There is nothing i can do in that case due to me not being able to disable this from code, but only from the shader, which i have no control over.

Again thanks go to: @sp614x for the help here.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,9 @@ repositories {
     maven {
         url 'http://dogforce-games.com/maven'
     }
+    flatDir {
+        dirs './optifine'
+    }
 }
 
 apply plugin: 'idea'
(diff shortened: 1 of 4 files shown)


src/main/java/com/structurize/compat/optifine/OptifineCompat.java
```diff
@@ -0,0 +1,188 @@
+package com.structurize.compat.optifine;
```
Raycoms, line 1: Can you please run our formatter on this one?

build.gradle
```diff
@@ -29,6 +29,9 @@ repositories {
     maven {
         url 'http://dogforce-games.com/maven'
     }
+    flatDir {
```
Raycoms: What does this do? Do we need this?

src/main/java/com/structurize/compat/optifine/OptifineCompat.java
marchermans: Did I not do that?

Raycoms: nope =D

PR #10, opened 2019-01-10
Feature/render more

Raycoms:
- Allow to render several things at the same time.
- Allow to add them to the same tessellator
-

Review please


cltnschlosser:
What was the problem with instances? Seems like a better approach than the Map with string identifiers.

marchermans (changes requested):
No.
Use a different tesselator,
We made them now instanced for a reason.

This just becomes more and more of an unmaintainable mess.

PR #12, opened 2019-01-11
CleanUp of the multi-template-rendering mess.

marchermans:
This cleans up the stuff @Raycoms and @someaddons patched in to make multi template rendering work.

This was the design all along, from when i first started working on this.
Adds compatibility for a lot of things as well: Entities, proper world handling etc.

Also rotates and mirror tile entities and entities now properly, which was never working in any version before.

Please review.

### DIFF
--- a/src/main/java/com/structurize/coremod/proxy/ClientProxy.java
+++ b/src/main/java/com/structurize/coremod/proxy/ClientProxy.java
@@ -218,12 +218,6 @@ public class ClientProxy extends CommonProxy
           (b) -> b.blockState.getBlock() instanceof BlockSubstitution,
           (b) -> new Template.BlockInfo(b.pos, Blocks.AIR.getDefaultState(), null)
         );
-
-        //We disable rendering of the Banner Block
-        TemplateBlockInfoTransformHandler.getInstance().AddTransformHandler(
-          blockInfo -> blockInfo.blockState.getBlock() instanceof BlockBanner,
-          blockInfo -> new Template.BlockInfo(blockInfo.pos, Blocks.AIR.getDefaultState(), null)
-        );
     }
 
     @Override
--- a/src/main/java/com/structurize/coremod/proxy/ClientProxy.java
+++ b/src/main/java/com/structurize/coremod/proxy/ClientProxy.java
@@ -16,6 +16,7 @@ import com.structurize.coremod.event.ClientEventHandler;
 import com.structurize.coremod.items.ModItems;
 import com.structurize.structures.client.TemplateBlockAccess;
 import com.structurize.structures.client.TemplateBlockInfoTransformHandler;
+import com.structurize.structures.client.TemplateEntityInfoTransformHandler;
 import com.structurize.structures.event.RenderEventHandler;
 import com.structurize.structures.helpers.Settings;
 import net.minecraft.block.Block;
@@ -28,6 +29,7 @@ import net.minecraft.client.renderer.block.statemap.StateMap;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.init.Blocks;
 import net.minecraft.item.Item;
+import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.stats.RecipeBook;
 import net.minecraft.util.ResourceLocation;
 import net.minecraft.util.math.BlockPos;
@@ -224,6 +226,8 @@ public class ClientProxy extends CommonProxy
           (b) -> b.blockState.getBlock() == ModBlocks.blockSolidSubstitution,
           (b) -> new Template.BlockInfo(b.pos, Blocks.AIR.getDefaultState(), null)
         );
+
+
     }
 
     @Override
--- a/src/main/java/com/structurize/coremod/proxy/ClientProxy.java
+++ b/src/main/java/com/structurize/coremod/proxy/ClientProxy.java
@@ -215,7 +215,13 @@ public class ClientProxy extends CommonProxy
 
         //Additionally we register an exclusion handler here;
         TemplateBlockInfoTransformHandler.getInstance().AddTransformHandler(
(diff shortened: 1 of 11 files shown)


src/main/java/com/structurize/coremod/proxy/ClientProxy.java
```diff
@@ -212,10 +214,16 @@ public static void registerModels(@NotNull final ModelRegistryEvent event)
         createCustomModel(ModBlocks.multiBlock);
 
         //Additionally we register an exclusion handler here;
-        TemplateBlockAccessTransformHandler.getInstance().AddTransformHandler(
+        TemplateBlockInfoTransformHandler.getInstance().AddTransformHandler(
           (b) -> b.blockState.getBlock() instanceof BlockSubstitution,
```
cltnschlosser: Are Minecraft blocks still singletons? Can you do a comparison over an instance of check.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
```diff
@@ -0,0 +1,158 @@
...
+
+        postTemplateDraw();
+    }
+
+    private static void preTemplateDraw(final Rotation rotation, final Mirror mirror, final Vector3d drawingOffset, final BlockPos inTemplateOffset)
+    {
+        final ITextureObject textureObject = Minecraft.getMinecraft().getTextureMapBlocks();
+        GlStateManager.bindTexture(textureObject.getGlTextureId());
+
+        GlStateManager.pushMatrix();
```
cltnschlosser, line 146: Do these get applied in the opposite order? I’d think you want to: scale, mirror, rotate, then translate.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
```diff
@@ -0,0 +1,158 @@
...
+        GlStateManager.translate(-rotateInTemplateOffset.getX(), -rotateInTemplateOffset.getY(), -rotateInTemplateOffset.getZ());
+
+        RenderUtil.applyRotationToYAxis(rotation);
+        RenderUtil.applyMirror(mirror, inTemplateOffset);
+
+        GlStateManager.scale(HALF_PERCENT_SHRINK, HALF_PERCENT_SHRINK, HALF_PERCENT_SHRINK);
+        GlStateManager.enableBlend();
+        GlStateManager.blendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+        GlStateManager.resetColor();
+        GlStateManager.pushMatrix();
```
cltnschlosser, line 159: Is this push needed? If yes. Is the first one needed. Seems weird to need 2.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
```diff
@@ -0,0 +1,158 @@
...
+    public void draw(final Rotation rotation, final Mirror mirror, final Vector3d drawingOffset, final float partialTicks, final BlockPos pos)
+    {
+        //Handle things like mirror, rotation and offset.
+        preTemplateDraw(rotation, mirror, drawingOffset, primaryBlockOffset);
+
+        //Draw normal blocks.
+        tessellator.draw();
+
+        //Draw tile entities.
+        tileEntities.forEach(tileEntity -> TileEntityRendererDispatcher.instance.render(tileEntity, partialTicks, 0));
```
cltnschlosser: Do you need to set positions for these? Or is that handled by the OpenGL translation too?

src/main/java/com/structurize/coremod/proxy/ClientProxy.java
marchermans: This works for all substitution blocks in one go. Else i would need two checks.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
marchermans: The term is TRSR : Translate Rotate Scale Rotate
In that order modifications need to be applied. Allthought it works a bit different here due to how mirroring works in our structures.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
marchermans: I always push twice to ensure a clean state when doing the actual rendering.
This does not cost any real performance.
And it is just cleaner todo.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
marchermans: The positions inside the template are handled during the construction of the TE.
The position of the template in the world is handled via the GL translate.

src/main/java/com/structurize/structures/client/TemplateRenderer.java
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Draws an instance of the template at the given position, with the given rotation, and mirroring.
+     *
+     * @param rotation The rotation.
+     * @param mirror The mirroring.
+     * @param drawingOffset The drawing offset.
+     * @param partialTicks The partial tick time.
+     * @param pos The position.
+     */
+    public void draw(final Rotation rotation, final Mirror mirror, final Vector3d drawingOffset, final float partialTicks, final BlockPos pos)
```
someaddons: since the position is done by the drawingOffset, is the final BlockPos pos parameter needed any longer?

marchermans: I am still testing and cleaning up of things that i do not need.
Right now i got most things rendering, but chests and for example the stand of an armor stand are note displaying.

src/main/java/com/structurize/coremod/proxy/ClientProxy.java
marchermans: @cltnschlosser I changed it.

Raycoms:
:shipit:

PR #16, opened 2019-01-26
Feature/blueprints

Raycoms:
- Finishes the blueprint support for @JTK222
- Updates formatting
- Removes not needed lines and classes
- Fixes up cash
- Fix up desert placeholders

Review please


marchermans (changes requested):
Codewise this is Okey. Before this gets merged I want to see documentation for this file format. 

I particular two things:
Exact description of how it works, what each property does, how to load it, etc. 
I need this for the web view I plan to extract information like required mods etc from it so I need to know that.

Also how did we save so much space?

Raycoms:
It is more efficient since it stores 
a) The blocks in a 3dim array and not the blocks with their blockpos
b) Doesn't store the blockstates etc but only an id of the state and puts the blockstate in a pallete

About doc, it has javadoc in all methods.

It has a getter for the modList for example and a getter for the architects (builders) of the style.
It is very easy to add more information to them as metadata if we need so.

All this info is in Blueprint.java

In BluePrintUtil it gets stored.

marchermans (approved):

PR #31, opened 2019-04-21
Fix buildtool view bounds not fitting the controls it contains

barthogenes:
- Change the view bounds of the build tool from this:
wrong bounds
to this:
right bounds

This fixes that you can click on the bottom half of the confirm and cancel buttons because they are within the bounds of the view now :)


Raycoms:
:shipit: just gotta sign the CLA

someaddons:
nice find^^

PR #59, opened 2019-07-27
Shingles && Data Generators

Asherslab:
- Adds new shingles variants
- Adds better Timber Frame blocks/models/variants
- Adds data generators for loot tables, recipes, blockstates, block models, item models, and etc.

Review please


Asherslab:
Don’t merge. Forgot to fix an issue

PR #82, opened 2019-11-05
Added cactus and planks to the BlockTags

RyTechro:
Review please

### DIFF
--- a/src/main/resources/data/minecraft/tags/blocks/logs.json
+++ b/src/main/resources/data/minecraft/tags/blocks/logs.json
@@ -1,0 +1,6 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:cactus"
+  ]
+}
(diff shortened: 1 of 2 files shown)


Raycoms:
did you test it?

RyTechro:
As good as I could. Started the game without the change and had no tag in the f3 menu and with this change they were there. Not sure how to test it otherwise, but seeing it in the f3 menu makes me confident that it worked

src/main/resources/data/minecraft/tags/blocks/logs.json
```diff
@@ -0,0 +1,6 @@
+{
+  "replace": false,
+  "values": [
+    "minecraft:cactus"
```
Nightenom, line 4: we definitely should not do that, probably won't break other mods but may cause weird things since cactus is usually not recognised as wood log

Raycoms: true, we might only want the plank tag

PR #91, opened 2020-02-08
Feature/faster rendering

Nightenom:
Review please


src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
```diff
@@ -78,6 +81,10 @@
      */
     private CompoundNBT[] entities = new CompoundNBT[0];
 
+    private List<BlockInfo> cacheBlockInfo = null;
+    private Map<BlockPos, BlockInfo> cacheBlockInfoMap = null;
```
Raycoms, line 88: javadoc

src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
```diff
@@ -331,21 +339,67 @@ public Blueprint setMissingMods(final String... missingMods)
...
+    public final Map<BlockPos, BlockInfo> getBlockInfoAsMap()
+    {
+        if (cacheBlockInfoMap == null)
+        {
+            getBlockInfoAsList0();
+        }
+        return cacheBlockInfoMap;
+    }
+
+    private final void getBlockInfoAsList0()
```
Raycoms: AsList0() ? method name and javadoc

src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
```diff
@@ -77,9 +77,17 @@ private BlueprintRenderer(final BlueprintBlockAccess blockAccess)
         init();
     }
 
+    public void updateBlueprint(final Blueprint blueprint)
```
Raycoms, line 87: jdoc

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
```diff
@@ -62,26 +52,49 @@ public static BlueprintHandler getInstance()
...
-            blueprintBufferBuilderCache.get(blueprint, () -> BlueprintRenderer.buildRendererForBlueprint(blueprint)).draw(pos, stack, partialTicks);
+            renderer = rendererRef.get();
         }
-        catch (ExecutionException e)
+
+        evictTimeCache.put(blueprintHash, now);
+        renderer.updateBlueprint(blueprint);
+        renderer.draw(pos, stack, partialTicks);
+
+        for (final Int2LongMap.Entry entry : evictTimeCache.int2LongEntrySet())
```
someaddons, line 90: I think this is probably not a good idea, to replicate cache-like behaviour without using a cache and looping over it each draw call, which are very high frequency. Probably gonna cause quite some fps loss.

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
```diff
@@ -62,26 +52,49 @@ public static BlueprintHandler getInstance()
...
             Log.getLogger().warn("Trying to draw null blueprint!");
             return;
         }
+        Minecraft.getInstance().getProfiler().startSection("struct_render_cache");
 
-        try
+        final int blueprintHash = blueprint.hashCode();
+        final SoftReference<BlueprintRenderer> rendererRef = rendererCache.get(blueprintHash);
+        final BlueprintRenderer renderer;
+        final long now = System.nanoTime();
```
someaddons: use System.miliseconds unless you really need nanoTime precision, nanoTime is relatively slow

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
Nightenom: The problem with old mechanism was that it simply hasn't worked as a cache (value was null more often than expected, tried some builder options without success), according to the old cache javadoc something like this is done every access so I doubt it's slower than the old one

Nightenom: Regarding fps loss, this entire patch is just for gigantic structures (more caching in blueprint class) and more lagless build tool usage (renderer refresh was called unnecessarily)

someaddons: the old one didnt have a timeout, just a maximum elements and caches aren't doing that stuff each access, else they'd be quite slow/unusable. Old one was a guava cache, see https://github.com/google/guava/wiki/CachesExplained#timed-eviction for how it works. Granted there are other caches out there aswell, some are faster. see here for example: https://cruftex.net/2016/03/16/Java-Caching-Benchmarks-2016-Part-1.html

Nightenom: as I said various builder options (incl. timed eviction) wasn't successful
I can probably connect this to everyGameSecond event instead of this one

someaddons: ye the other option to keep this would be to do it on a tick event, like simply subscribing to the client tick event and evaluate the times there. Should still be a lot better than here

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
```diff
@@ -62,25 +52,53 @@ public static BlueprintHandler getInstance()
...
+    public void cleanCache()
+    {
+        for (final Int2LongMap.Entry entry : evictTimeCache.int2LongEntrySet())
         {
-            Log.getLogger().error(e);
+            if (entry.getLongValue() + CACHE_EVICT_TIME < System.currentTimeMillis())
+            {
+                final int removeHash = entry.getIntKey();
+                final SoftReference<BlueprintRenderer> removeRendererRef = rendererCache.remove(removeHash);
+                evictTimeCache.remove(removeHash);
```
someaddons, line 96: can you actually remove it during iteration here without an exception? might want to user an Iterator isntead

src/main/java/com/ldtteam/structures/lib/BlueprintUtils.java
```diff
@@ -21,29 +16,21 @@
...
  * Utility functions for blueprints.
  */
 public final class BlueprintUtils
 {
-    private static final Cache<Blueprint, Map<BlockPos, BlockInfo>> blueprintBlockInfoCache = CacheBuilder.newBuilder().maximumSize(50).build();
-
-    private static final Set<String> blackListedTileEntityIds = Sets.newHashSet();
-    private static final Set<String> blackListedEntityIds = Sets.newHashSet();
+    // private static final Set<String> blackListedTileEntityIds = new HashSet<>();
+    // private static final Set<String> blackListedEntityIds = new HashSet<>();
```
someaddons, line 28: commented out?

Nightenom: It was there for reason which is now not correct. Leaving it here if it will return in future

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
Nightenom: its ArrayMap, and it never threw an exception during testing (was logging it so it surely happened many times)

Nightenom: btw the for each uses set method which creates a copy, so that's probably why

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
```diff
@@ -21,20 +20,11 @@
      * A static instance on the client.
      */
     private static final BlueprintHandler ourInstance = new BlueprintHandler();
+    private static final int CACHE_SIZE = 30;
```
marchermans, line 23: This can be done much easier: Use a guava cache.

src/main/java/com/ldtteam/structures/client/BlueprintBlockAccess.java
```diff
@@ -70,6 +70,11 @@ public Blueprint getBlueprint()
         return blueprint;
     }
 
+    public void setBlueprint(final Blueprint blueprintIn)
```
marchermans, line 73: A renderer should exist for a single Blueprint. And a renderer should not change its blueprint!

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
marchermans: I agree here -> This needs to change. Use a proper Guava Cache!

src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
marchermans: Just plain nope -> One renderer per Blueprint, is architecturally way better. Recreate it if you need to and evict the renderer from your cache!

src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
```diff
@@ -135,15 +148,13 @@ private void init()
      */
     public void draw(final BlockPos pos, final MatrixStack matrixStack, final float partialTicks)
     {
+        Minecraft.getInstance().getProfiler().startSection("struct_render_init");
```
marchermans, line 154: Good!

src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
```diff
@@ -135,15 +148,13 @@ private void init()
      */
     public void draw(final BlockPos pos, final MatrixStack matrixStack, final float partialTicks)
```
marchermans, line 152: This is recreating the buffer every tick!
Please rework this so that it creates one buffer per use case and reuses this.

What ever a use case is might then be up to you (Blueprint + pos or as in the original only blueprint) However this is not a nice way of doing things and a lot of performance is lost here in this method!
So if this PR is targeting performance please update this method before merging, since it will likely cause structural changes in a lot of other places.

src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
```diff
@@ -134,4 +141,19 @@ private static void renderBox(final BlockPos posA, final BlockPos posB, final Cl
...
+     * Call renderer cache cleaning every 5 secs (100 ticks).
+     *
+     * @param event the catched event.
+     */
+    @SubscribeEvent
+    public static void onClientTickEvent(final ClientTickEvent event)
+    {
+        if (Minecraft.getInstance().world.getGameTime() % (Constants.TICKS_SECOND * 5) == 0)
+        {
+            BlueprintHandler.getInstance().cleanCache();
```
marchermans, line 183: Nope! This is way to much of a CPU hog, use a Guava cache which can do this for you automatically!

marchermans (changes requested):
This only targets the low hanging fruit, and makes some very questionable choices when it comes to the way things are cached and handled. Additionally it cleans things up which are currently still under development : BlueprintTessellator.

Please rework this to target the largest slowdown compared to the old system first -> Rebuilding the rendering buffer every draw call. Then rework your caching logic by using a Guava cache which will reduce the complexity of this PR by 100-fold. After that please also take a look at the fact that you now have  a Renderer for a Blueprint (BlueprintRenderer) however in reality you are now rendering more then one blueprint with said renderer which is not a good design!

src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
Nightenom: Which buffer? I copied this from vanilla code. init() method is called when you open uncached schematic or rotate/mirror is called (that's not every tick)

PR #94, opened 2020-02-10
Zoom drag view, event propagation, int to double mouse conversion

Nightenom:
- Introduce new Zoom Drag view, user can zoom in/out and drag content like in most online maps. Rotation can be added into this view easily if anyone wanted
- Mouse events are now propagated using boolean return value and propagation is stopped if any pane returns true
- Mouse events now use double instead of int type for mouse position (source vanilla)

Review please


src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -46,7 +46,7 @@
         register("dropdown", DropDownList.class);
         register("overlay", OverlayView.class);
         register("gradient", Gradient.class);
-        register("dragView", DragView.class);
+        register("zoomdragview", ZoomDragView.class);
     }
```
Raycoms, line 50: we got both types in blockOut.xsd, but only one here, with a different name?

src/main/java/com/ldtteam/blockout/controls/Scrollbar.java
```diff
@@ -144,17 +144,17 @@ public void drawSelf(final int mx, final int my)
     }
 
     @Override
-    public void handleClick(final int mx, final int my)
+    public boolean handleClick(final double mx, final double my)
     {
         if (getContentHeightDiff() <= 0)
         {
-            return;
+            return true;
```
Nightenom: Not sure about return value. Does that mean scrollbar is not visible - should we then allow clicking in lower layers under scrollbar?

src/main/java/com/ldtteam/blockout/views/View.java
```diff
@@ -83,28 +82,26 @@ public void drawSelf(final int mx, final int my)
...
-        {
-            if (child != null)
-            {
-                child.scrollInput(wheel);
-            }
-        }
+        final int mxChild = (int) mx - x - padding;
+        final int myChild = (int) my - y - padding;
+
+        return reverseChildIterationStream().filter(child -> child.isPointInPane(mxChild, myChild))
```
Raycoms: I think all the streams you're opening here are not worth it, we never got that many elements to justify the setup overhead of a stream.

src/main/resources/assets/structurize/gui/windowbuildtool.xml
```diff
@@ -44,7 +44,7 @@
         <buttonimage id="plus" source="structurize:textures/gui/buildtool/plus.png" size="32" pos="64 64"/>
         <buttonimage id="confirm" source="structurize:textures/gui/buildtool/confirm.png" size="32" pos="0 96"/>
         <buttonimage id="cancel" source="structurize:textures/gui/buildtool/cancel.png" size="32" pos="64 96"/>
-    </view>
+    </zoomdragview>
```
Raycoms: gotta remove this from the buildtool window yet

src/main/java/com/ldtteam/blockout/controls/Scrollbar.java
Raycoms: Yes, in this case there is no scrollbar, so we should allow clicks at that location (either to drag, or for a button, or for whatever).

src/main/java/com/ldtteam/blockout/views/View.java
Nightenom: Suggests or create a better abstraction for this problem, I'm aware of streams being a bit slower than for-each loops. Mouse moving events might be a problem using this.

Raycoms: I'd just use normal loops yes.

src/main/java/com/ldtteam/blockout/BOScreen.java
```diff
@@ -63,14 +62,13 @@ public void render(final int mx, final int my, final float f)
...
-    public boolean keyPressed(final int key, final int p_keyPressed_2_, final int p_keyPressed_3_)
+    public boolean keyPressed(final int key, final int scanCode, final int modifiers)
     {
-        if (key == GLFW.GLFW_KEY_BACKSPACE || key == GLFW.GLFW_KEY_ESCAPE || key == GLFW.GLFW_KEY_ENTER || key == GLFW.GLFW_KEY_TAB || key == GLFW.GLFW_KEY_BACKSPACE ||
-            key == GLFW.GLFW_KEY_INSERT || key == GLFW.GLFW_KEY_DELETE || key == GLFW.GLFW_KEY_RIGHT || key == GLFW.GLFW_KEY_LEFT || key == GLFW.GLFW_KEY_DOWN ||
-            key == GLFW.GLFW_KEY_UP || key == GLFW.GLFW_KEY_PAGE_UP || key == GLFW.GLFW_KEY_PAGE_DOWN || key == GLFW.GLFW_KEY_HOME || key == GLFW.GLFW_KEY_END ||
-            // key == GLFW.GLFW_KEY_CAPS_LOCK ||
-            // key == GLFW.GLFW_KEY_SCROLL_LOCK ||
-            // key == GLFW.GLFW_KEY_NUM_LOCK ||
+        if (key == GLFW.GLFW_KEY_BACKSPACE || key == GLFW.GLFW_KEY_ESCAPE || key == GLFW.GLFW_KEY_ENTER || key == GLFW.GLFW_KEY_TAB ||
```
someaddons: I know its not new, but think it might be a good idea to just put all those keys into a hashmap and then just check contains?

Nightenom:
Also improved zooming to be more user-friendly

src/main/java/com/ldtteam/blockout/Pane.java
```diff
@@ -249,14 +249,6 @@ public void disable()
         setEnabled(false);
     }
 
-    /**
-     * Is pane visible and enabled.
-     */
-    public boolean isOn()
```
someaddons, line 255: hm is this used nowhere in minecolonies either?

someaddons (approved):
looks good

src/main/java/com/ldtteam/blockout/Pane.java
Nightenom: I added it before year approximately so I doubt anyone used it. Whatever you shouldn't need to ever use it

PR #102, opened 2020-03-11
Tree view

Nightenom:
- adds tree view
- panes are backed using nodes - every node contains its childs and updater
- the updater can be disabled if direct access to pane is better for accessing
- TreeView#reconstruct() builds only missing panes, you need to call this method to reposition panes in entire tree

Review please
Don't merge before test classes are reverted


marchermans (approved):

Raycoms:
:shipit:

PR #114, opened 2020-04-06
Added lineWidth setter.

Kosmic138:
- added setter method for property lineWidth of class Box.


Asherslab:
Out of curiosity, why is this wanted/needed?

Things like that should generally be defined by the XML sheets, not a setter like that

Kosmic138:
I am updating townHall information in MineColonies. I am creating dynamic layout for list of guard building and their respective guards. There is already a color setter in that class so it makes sense for me to have the second setter as well.

Asherslab (approved):
That's reasonable enough for me.

PR #107, opened 2020-03-22
Model Optimizations

WenXin20:
Closes ldtteam/minecolonies#4139

# Changes proposed in this pull request:
- Optimizes Multiblock model
- Fixes item display properties
- Updates the placeholder & substitution blocks to the 1.14+ texture for planks

Review please

PR #117, opened 2020-04-09
Fix Data Generators, Run generators, And add Floating Carpets!

Asherslab:
Ignore the JSON files if you know what's good for you.
- Fixes the generated JSON stuff up from #107 
- Adds Floating Carpets, literally normal carpets  but they *don't* need support underneath them (no more visible string to have carpets!)


Nightenom (approved):

Asherslab:
Approved by ray

PR #119, opened 2020-04-11
fix memory leak in rendering

Nightenom:
- fix memory leak caused by vertex buffers

Review please


marchermans (changes requested):
This is still leaving the buffers open when the user changes rotation or mirroring of a schematic!

marchermans (changes requested):
No this needs to happen at the top. Else u ou delete the data from the gpu again.

Nightenom:
> No this needs to happen at the top. Else u ou delete the data from the gpu again.

Hmm?

marchermans (approved):

PR #128, opened 2020-05-01
Fix water handling of the building tool

ToMe25:
- Fix how the building tool paste button handles water
- Fix how the building tool replace block function handles water
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
@@ -1,13 +1,14 @@
 package com.ldtteam.structurize.util;
 
-import com.ldtteam.structurize.api.util.ItemStackUtils;
 import com.ldtteam.structurize.blocks.decorative.BlockTimberFrame;
 import net.minecraft.block.*;
 import net.minecraft.fluid.Fluids;
 import net.minecraft.item.*;
 import net.minecraft.state.BooleanProperty;
 import net.minecraft.tags.ItemTags;
-import net.minecraft.util.*;
+import net.minecraft.util.Direction;
+import net.minecraft.util.Hand;
+import net.minecraft.util.Rotation;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.BlockRayTraceResult;
 import net.minecraft.util.math.Vec3d;
@@ -401,7 +402,7 @@ public final class BlockUtils
         final Block block = state.getBlock();
         if((!(block instanceof IBucketPickupHandler) || ((IBucketPickupHandler)block).pickupFluid(world, pos, state) == Fluids.EMPTY) && block instanceof FlowingFluidBlock)
         {
-            world.setBlockState(pos, Blocks.AIR.getDefaultState(), 3);
+            world.setBlockState(pos, Blocks.AIR.getDefaultState(), com.ldtteam.structurize.api.util.constant.Constants.UPDATE_FLAG);
         }
     }
 }
--- a/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
@@ -397,10 +397,7 @@ public final class BlockUtils
     {
         final BlockState state = world.getBlockState(pos);
         final Block block = state.getBlock();
-        if (block instanceof IBucketPickupHandler && ((IBucketPickupHandler)block).pickupFluid(world, pos, state) != Fluids.EMPTY)
-        {
-        }
-        else if (block instanceof FlowingFluidBlock)
+        if((!(block instanceof IBucketPickupHandler) || ((IBucketPickupHandler)block).pickupFluid(world, pos, state) == Fluids.EMPTY) && block instanceof FlowingFluidBlock)
         {
             world.setBlockState(pos, Blocks.AIR.getDefaultState(), 3);
         }
--- a/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
@@ -200,12 +200,14 @@ public final class BlockUtils
         }
         else if (blockState.getBlock() instanceof StemBlock)
         {
-            final ItemStack stack = ((StemBlock) blockState.getBlock()).getItem(null, null, blockState);
-            if (!ItemStackUtils.isEmpty(stack))
+            if (blockState.getBlock() == Blocks.PUMPKIN)
             {
-                return stack.getItem();
+                return Items.PUMPKIN_SEEDS;
             }
-            return Items.MELON_SEEDS;
+            else
+            {
+                return blockState.getBlock()== Blocks.MELON ? Items.MELON_SEEDS : null;
+            }
         }
         else if (blockState.getBlock() == Blocks.BAMBOO_SAPLING)
         {
--- a/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
@@ -3,6 +3,7 @@ package com.ldtteam.structurize.util;
 import com.ldtteam.structurize.api.util.ItemStackUtils;
 import com.ldtteam.structurize.blocks.decorative.BlockTimberFrame;
 import net.minecraft.block.*;
+import net.minecraft.fluid.Fluids;
 import net.minecraft.item.*;
 import net.minecraft.state.BooleanProperty;
 import net.minecraft.tags.ItemTags;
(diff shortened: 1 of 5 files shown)


src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -386,4 +386,23 @@ else if (stackToPlace.getItem() == Items.WATER_BUCKET)
...
+     * 
+     * @param world the world to remove the fluid from.
+     * @param pos   the position where to remove the fluid.
+     */
+    public static void removeFluid(World world, BlockPos pos)
+    {
+        final BlockState state = world.getBlockState(pos);
+        final Block block = state.getBlock();
+        if (block instanceof IBucketPickupHandler && ((IBucketPickupHandler)block).pickupFluid(world, pos, state) != Fluids.EMPTY)
+        {
```
Raycoms: empty if?

ToMe25: the block is getting emptied in the if, the if is just needed to detect whether it worked.

ToMe25: as it doesn't work with non-source blocks, but they should get removed too, there has to still be the else, so i just have an empty if there.

ToMe25:
This is probably finished now

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -386,4 +388,20 @@ else if (stackToPlace.getItem() == Items.WATER_BUCKET)
...
+     * @param world the world to remove the fluid from.
+     * @param pos   the position where to remove the fluid.
+     */
+    public static void removeFluid(World world, BlockPos pos)
+    {
+        final BlockState state = world.getBlockState(pos);
+        final Block block = state.getBlock();
+        if((!(block instanceof IBucketPickupHandler) || ((IBucketPickupHandler)block).pickupFluid(world, pos, state) == Fluids.EMPTY) && block instanceof FlowingFluidBlock)
+        {
+            world.setBlockState(pos, Blocks.AIR.getDefaultState(), 3);
```
Nightenom, line 404: use Constants.UPDATE_FLAG instead 3

Nightenom: use Constants.UPDATE_FLAG instead 3

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
```diff
@@ -94,63 +128,81 @@ public void setupStructurePlacement(@NotNull final BlockPos pos, final boolean c
...
      * @return the last pos.
      */
     public BlockPos placeStructure(final World world, final ChangeStorage storage, final BlockPos inputPos)
+    {
+        return placeStructure(world, storage, inputPos, (structure, pos) -> structure.getBlockState(pos).getBlock() == ModBlocks.blockSubstitution
+                                                                           || structure.getBlockState(pos).getBlock() instanceof IAnchorBlock);
+    }
+
+    /**
+     *
```
Nightenom, line 146: Javadoc

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
```diff
@@ -94,63 +128,81 @@ public void setupStructurePlacement(@NotNull final BlockPos pos, final boolean c
...
     {
         structure.setLocalPosition(inputPos);
         @NotNull final List<BlockPos> delayedBlocks = new ArrayList<>();
         final BlockPos endPos = new BlockPos(this.structure.getWidth(), this.structure.getHeight(), this.structure.getLength());
-        BlockPos currentPos = inputPos;
+        BlockPos currentPos = new BlockPos(inputPos.getX(), inputPos.equals(BlockPos.ZERO) ? endPos.getY() - 1 : inputPos.getY(), inputPos.getZ());
         int count = 0;
 
-        for (int y = currentPos.getY(); y < endPos.getY(); y++)
+        for (int y = currentPos.getY(); y >= 0; y--)
```
Nightenom: You can't iterate from up to down because of falling blocks, or make them go into delayed blocks (and even that is not a complete solution)

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
```diff
@@ -94,63 +128,81 @@ public void setupStructurePlacement(@NotNull final BlockPos pos, final boolean c
...
                     else
                     {
+                        BlockUtils.removeFluid(world, worldPos);
                         delayedBlocks.add(localPos);
                     }
 
                     if (count >= Structurize.getConfig().getCommon().maxOperationsPerTick.get())
                     {
-                        this.handleDelayedBlocks(delayedBlocks, storage, world);
+                        handleDelayedBlocks(delayedBlocks, storage, world);
```
Nightenom, line 227: Are you sure you want to call this every block placement?

Nightenom (changes requested):
Rather than reverse iterating, we should remove fluids in a single tick before the structure placement starts. Since removing a single block of water is usually not sufficient to remove entire water source.

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
Nightenom: My bad, misread git diff

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
ToMe25: done

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
ToMe25: done

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
ToMe25: how does iterating from the top cause any new problems with falling blocks?
it shouldn't create any the from that bottom up wouldn't have created, and both would ignore falling blocks after that.

Nightenom: when you place a falling block and it has no support it falls down, and most of falling blocks are solid so they don't go into the delayed blocks

ToMe25: oh, now i understand, yes thats a problem, i thought with blocks that should get broken and can fall.

ToMe25: what about just adding all the fluid blocks to remove to delayedBlocks?
That all gets handled in a single tick, so the order wouldn't be a problem for fluids.

ToMe25: oops, they don't get handled all in the same tick.

Nightenom: you would have to prioritize them because delayedBlocks are mostly non-solid blocks, also now imagine you have lava block in the structure and it becomes obsidian.
btw this placing mechanism doesn't cover many problematic situations.
I would say the best solution which does not mess up other things is to simply clear all fluids that do not match the structure in a single tick as the first step

ToMe25: yes, thats probably right, i thought all the delayed blocks would get handled in a single tick, then it might have worked, but not like this.

ToMe25: but i don't think that with the obsidian would happed anyways.

ToMe25: I did that, i am just not sure whether its a good idea to only remove fluids if the input pos is 0,0,0, or if that will just cause strange bugs again.

Nightenom: You can actually iterate from up to down in while removing fluid, but don't do that when placing

Nightenom: and rares cases are still cases :P

ToMe25: > You can actually iterate from up to down in while removing fluid, but don't do that when placing

yes, but it doesn't matter.
as this is all in a single tick, none of the fluids can flow, so it can be ignored.

> and rares cases are still cases :P

yes, but i did some tests and the overall performance was better without that check, as he has to check that way more often, then he could actually keep the fluid.

Nightenom (approved):
If @ToMe25 thinks it's ready then merge it

ToMe25:
i think it is ready, and i tested it somewhat, however i am not sure whether it might still have some hidden issues.

PR #137, opened 2020-05-17
Placement rework

Raycoms:
Mayor overhaul of placement handling:

Remove "Structure" class.

Add new StructureIterator class which is used to iterate over a structure.
Add StructureHandler interface and simple implementation to handle structure placement with nice callbacks

-
-
-

Review please


src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
```diff
@@ -357,7 +365,57 @@ public Blueprint setMissingMods(final String... missingMods)
         return cacheBlockInfoMap;
     }
 
-    private final void buildBlockInfoCaches()
+    /**
+     * Getter of the EntityInfo at a certain position.
+     *
+     * @param pos the position.
+     * @param structurePos the position it will have in the structure.
+     * @return the blockState.
```
someaddons: returns TE compound with real world coords

src/main/java/com/ldtteam/structures/helpers/AbstractStructureHandler.java
```diff
@@ -0,0 +1,116 @@
...
+     * Abstract constructor of structure handler.
+     * @param world the world it gets.
+     * @param pos the position the anchor of the structure got placed.
+     * @param structureName the name of the structure.
+     * @param settings the placement settings.
+     */
+    public AbstractStructureHandler(final World world, final BlockPos pos, final String structureName, final PlacementSettings settings)
+    {
+        this.world = world;
+        this.position = pos;
```
someaddons: can we rename position to worldPosition? since we handle both real world locations and relative schematic positions I'd like if we could identify that by name

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+    private final BlockPos size;
+
+    /**
+     * The structure handler.
+     */
+    private final IStructureHandler structureHandler;
+
+    /**
+     * Initialize the blueprint iterator with the structure handler.
+     * @param structureHandler the size of the blueprint.
```
someaddons: outdated javadoc

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+                    this.reset();
+                    return false;
+                }
+            }
+        }
+
+        return true;
+    }
+
+    public boolean increment(final Predicate<BlueprintPositionInfo> blockInfoPredicate, final boolean includeEntities)
```
someaddons: javadoc missing

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+            else if (BlockUtils.areBlockStatesEqual(structureHandler.getBluePrint().getBlockInfoAsMap().get(progressPos).getState(), structureHandler.getWorld().getBlockState(worldPos)))
+            {
+                structureHandler.triggerSuccess(progressPos);
+            }
+            else if (!blockInfoPredicate.test(info))
+            {
+                return true;
+            }
+        }
+        while (count < Structurize.getConfig().getCommon().maxBlocksChecked.get());
```
someaddons: how about we let the handler decide how many blocks to do at once?

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+
+        return true;
+    }
+
+
+    public boolean decrement(final Predicate<BlueprintPositionInfo> blockInfoPredicate, final boolean includeEntities)
```
someaddons: javadoc

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+            else if (BlockUtils.areBlockStatesEqual(structureHandler.getBluePrint().getBlockInfoAsMap().get(progressPos).getState(), structureHandler.getWorld().getBlockState(worldPos)))
+            {
+                structureHandler.triggerSuccess(progressPos);
+            }
+            else if (!blockInfoPredicate.test(info))
+            {
+                return true;
+            }
+        }
+        while (count < Structurize.getConfig().getCommon().maxBlocksChecked.get());
```
someaddons: same as above, let handler determine amount of steps/blocks at once

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+    private final IStructureHandler structureHandler;
+
+    /**
+     * Initialize the blueprint iterator with the structure handler.
+     * @param structureHandler the size of the blueprint.
+     */
+    public BlueprintIterator(final IStructureHandler structureHandler)
+    {
+        this.structureHandler = structureHandler;
+        this.size = new BlockPos(structureHandler.getBluePrint().getSizeX(), structureHandler.getBluePrint().getSizeY(), structureHandler.getBluePrint().getSizeZ());
```
someaddons: can we let the placement handler determine the size we iterate? could do partial/combined building of multiple handlers that way

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,178 @@
...
+        return true;
+    }
+
+
+    public boolean decrement(final Predicate<BlueprintPositionInfo> blockInfoPredicate, final boolean includeEntities)
+    {
+        int count = 0;
+        do
+        {
+            final BlockPos worldPos = structureHandler.getPosition().subtract(structureHandler.getBluePrint().getPrimaryBlockOffset()).add(progressPos);
```
someaddons: isnt structureHandler.getPosition() our world pos of the primary offset(anchor) already?

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -220,6 +223,49 @@ else if (blockState.getBlock() == Blocks.BAMBOO_SAPLING)
...
+     * For structure placement, check if two blocks are alike or if action has to be taken.
+     * @param blockState1 the first blockState.
+     * @param blockState2 the second blockState.
+     * @return true if nothing has to be done.
+     */
+    public static boolean areBlockStatesEqual(final BlockState blockState1, final BlockState blockState2)
+    {
+        if (blockState2 == null)
+        {
+            return true;
```
someaddons, line 239: equals returns true if the second is null? that seems weird/different from what I would expect it to return

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -220,6 +223,49 @@ else if (blockState.getBlock() == Blocks.BAMBOO_SAPLING)
...
+        final Block structureBlock = blockState2.getBlock();
+        final Block worldBlock = blockState1.getBlock();
+
+        if (structureBlock == ModBlocks.blockSubstitution || blockState2.equals(blockState1))
+        {
+            return true;
+        }
+
+        if (structureBlock == ModBlocks.blockSolidSubstitution)
+        {
```
someaddons: wouldnt this sth we can move into the handler? instead of checking it before

src/main/java/com/ldtteam/structurize/util/BlueprintPositionInfo.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Contains all information we need for placement at a given position.
+ */
+public class BlueprintPositionInfo
+{
+    /**
+     * The position in the structure.
+     */
+    private final BlockPos pos;
```
someaddons, line 16: this is a relative schematic position right?

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
someaddons: or maybe its better to just put that into the constructor parameters

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Nightenom: This should be based on server tick length, there is similar logic for timing in rendering pipeline which provides "stop" break when the remaining time is low.
Count not incremented?

src/main/java/com/ldtteam/structurize/Structurize.java
```diff
@@ -42,6 +44,8 @@ public Structurize()
         Mod.EventBusSubscriber.Bus.MOD.bus().get().register(LifecycleSubscriber.class);
         Mod.EventBusSubscriber.Bus.FORGE.bus().get().register(EventSubscriber.class);
         DistExecutor.runWhenOn(Dist.CLIENT, () -> () -> Mod.EventBusSubscriber.Bus.FORGE.bus().get().register(ClientEventSubscriber.class));
+
+        BlockSolidSubstitution.NOT_SOLID.add(blockState -> !blockState.getMaterial().isSolid() || blockState.getBlock() instanceof AirBlock);
```
Nightenom: Probably move this to proper loading phase, rather than mod constructor

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -220,6 +223,49 @@ else if (blockState.getBlock() == Blocks.BAMBOO_SAPLING)
...
+
+        //list of things to only check block for.
+        //For the time being any flower pot is equal to each other.
+        if (structureBlock instanceof DoorBlock || structureBlock == Blocks.FLOWER_POT)
+        {
+            return structureBlock == worldBlock;
+        }
+
+        return (structureBlock == Blocks.GRASS_BLOCK && worldBlock == Blocks.DIRT)
+                 || worldBlock == Blocks.GRASS_BLOCK && structureBlock == Blocks.DIRT;
```
Nightenom: This is definitely at the wrong level, it should be handled by minecolonies since it's builder's problem and not placement thing

src/main/java/com/ldtteam/structurize/util/InstantStructurePlacer.java
```diff
@@ -62,18 +70,18 @@ public InstantStructurePlacer(final Structure structure)
...
-        structure.setLocalPosition(pos);
-        for (int j = 0; j < this.structure.getHeight(); j++)
+        iterator.setProgressPos(new BlockPos(0, 0, 0));
+        for (int j = 0; j < this.handler.getBluePrint().getSizeY(); j++)
         {
-            for (int k = 0; k < this.structure.getLength(); k++)
+            for (int k = 0; k < this.handler.getBluePrint().getSizeZ(); k++)
             {
-                for (int i = 0; i < this.structure.getWidth(); i++)
+                for (int i = 0; i < this.handler.getBluePrint().getSizeZ(); i++)
```
Nightenom: getSizeX

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -220,6 +223,49 @@ else if (blockState.getBlock() == Blocks.BAMBOO_SAPLING)
...
+                }
+            }
+        }
+
+        //list of things to only check block for.
+        //For the time being any flower pot is equal to each other.
+        if (structureBlock instanceof DoorBlock || structureBlock == Blocks.FLOWER_POT)
+        {
+            return structureBlock == worldBlock;
+        }
```
Nightenom: If I'm not blind, this entire function is missing `structureBlock == worldBlock` check
Probably also remove this if?

Raycoms: On purpose. However, I think we can remove the door part in 1.15 too. Although flowerpots we probably want to handle like this still.

src/main/java/com/ldtteam/structurize/Structurize.java
Raycoms: To which one would you move this?

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
Raycoms: True, I'll add something to facilitate this.

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Raycoms: Partial combined handling I would always handling by cutting it into multiple blueprints rather than giving a partial blueprint (because it would make iteration less complex).

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Raycoms: Yeah, I'm not completely finished on this part yet.

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Raycoms: It is. So worldPos.subtract(primaroffset) = 0,0,0 pos of schematic + add progress = pos

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
Raycoms: I'll add it for both, but if blockstate2 is null, this means that we're likely outside of the schematic in this call somehow, so we want to handle this smoothly.

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
someaddons: but we could simply allow partial iteration here

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
someaddons: ah makes sense

src/main/java/com/ldtteam/structurize/util/BlueprintPositionInfo.java
Raycoms: yes

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
Raycoms: Then we have to handle all of them, and we don't want to have to handle them.

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Raycoms: Added a second constructor so both is possible.

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
Nightenom: there is Wold#isOutsideBuildHeight or isOutsideY or however it's called, worth using that

src/main/java/com/ldtteam/structurize/Structurize.java
Nightenom: I don't know :D probably block registration since registries are before common setup, or even static block code of ModBlocks class

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
Raycoms: I mean it's when the blueprint returns null for a position.

Nightenom: blueprint cant return null for pos, if it does then it's corrupted

Nightenom: and how do you get outside of the schematic?

Raycoms: It can, if out of range. (-1 pos for example)

src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
```diff
@@ -700,15 +763,50 @@ public boolean equals(final Object obj)
...
+     * @param pos the position
+     * @param includeEntities if entities should be included.
+     * @return the info object.
+     */
+    public BlueprintPositionInfo getBluePrintPositionInfo(final BlockPos pos, final boolean includeEntities)
+    {
+        return new BlueprintPositionInfo(pos, getBlockInfoAsMap().get(pos),
+          includeEntities
+            ? Arrays.stream(this.getEntities()).filter(data -> data != null && isAtPos(data, pos)).toArray(CompoundNBT[]::new)
+            : new CompoundNBT[0]);
```
someaddons: depending on how often this is called I think we could end up with n * n searches? maybe make a HashMap<Pos,EntityData> the first time it wants entity data

src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
```diff
@@ -700,15 +763,50 @@ public boolean equals(final Object obj)
...
+
+    /**
+     * Check if an entityData object is at the local position.
+     * @param entityData the data object to check.
+     * @param pos the pos to check.
+     * @return true if so.
+     */
+    private boolean isAtPos(@NotNull final CompoundNBT entityData, final BlockPos pos)
+    {
+        final ListNBT list = entityData.getList("Pos", 6);
```
someaddons: "Pos" should probably be a tag constant, think we use it elsewhere too

src/main/java/com/ldtteam/structures/helpers/AbstractStructureHandler.java
```diff
@@ -0,0 +1,122 @@
+package com.ldtteam.structures.helpers;
```
someaddons: can we refactor the package from helpers to placement? think that'd fit better, I'd normally expect utilities in helpers :D

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
```diff
@@ -0,0 +1,234 @@
...
+        this.includeEntities = true;
+    }
+
+    /**
+     * Reset the progressPos.
+     */
+    public void reset()
+    {
+        BlockPosUtil.set(this.progressPos, NULL_POS);
+        includeEntities = false;
```
someaddons: I think resetting the includeEntities flag here is misleading, should just keep whatever it was set to

src/main/java/com/ldtteam/structures/helpers/CreativeStructureHandler.java
```diff
@@ -0,0 +1,112 @@
...
+import com.ldtteam.structurize.util.PlacementSettings;
+import net.minecraft.block.BlockState;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import net.minecraftforge.items.IItemHandler;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A handler for structures. To place a structure a handler is required.
```
someaddons: too generic javadoc :D

src/main/java/com/ldtteam/structures/helpers/IStructureHandler.java
```diff
@@ -0,0 +1,208 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+
+//todo add to minecolonies ores to solidSubStitutionPredicateList
+//todo Minecolonies implement this with the AI.
+// return (block1 == Blocks.GRASS_BLOCK && block2 == Blocks.DIRT) || block2 == Blocks.GRASS_BLOCK && block1 == Blocks.DIRT;
+
+/**
+ * A handler for structures. To place a structure a handler is required.
```
someaddons: could you add 2-3 simple sentences about what task a StructureHandler does here as javadoc?

src/main/java/com/ldtteam/structures/helpers/SurvivalStructureHandler.java
```diff
@@ -0,0 +1,105 @@
...
+import com.ldtteam.structurize.util.PlacementSettings;
+import net.minecraft.block.BlockState;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+import net.minecraftforge.items.IItemHandler;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A handler for structures. To place a structure a handler is required.
```
someaddons: copy paste javadoc ^^

src/main/java/com/ldtteam/structures/helpers/SurvivalStructureHandler.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class SurvivalStructureHandler extends AbstractStructureHandler
+{
+    private final IItemHandler inv;
```
someaddons: missing javadoc

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -0,0 +1,358 @@
...
+ * if result = missingItems
+ * return MISSING_ITE
+ * else if result = block break
+ * return BREAK_BLOCK
+ * else if result = finished
+ * return nextState
+ * else
+ * return getState()
+ *
+ * on success we have to write the increased progress pos to NBT.
```
someaddons: what is this? xD

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -0,0 +1,358 @@
...
+ */
+
+/**
+ * Interface for using the structure codebase.
```
someaddons: this isnt an interface^^

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -0,0 +1,358 @@
...
+            final BlockPos worldPos = handler.getWorldPos().add(localPos).subtract(handler.getBluePrint().getPrimaryBlockOffset());
+            if (storage != null)
+            {
+                storage.addPositionStorage(worldPos, world);
+            }
+
+            final BlockPlacementResult result;
+            switch (operation)
+            {
+                case BLOCK_REMOVAL:
```
someaddons, line 113: hm wondering about having this whole thing within the structure handler, so it determines which action it wants for given blueprint data, pos

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -0,0 +1,358 @@
...
+
+        if (!(worldState.getBlock() instanceof AirBlock))
+        {
+            if (!handler.allowReplace())
+            {
+                return new BlockPlacementResult(worldPos, BlockPlacementResult.Result.BREAK_BLOCK);
+            }
+        }
+
+        for (final CompoundNBT compound : iterator.getBluePrintPositionInfo().getEntities())
```
someaddons: shouldn't this depend on includeEntities?

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -220,6 +230,58 @@ else if (blockState.getBlock() == Blocks.BAMBOO_SAPLING)
...
+            {
+                if (notSolid.test(blockState1))
+                {
+                    return true;
+                }
+            }
+
+            //list of things to only check block for.
+            //For the time being any flower pot is equal to each other.
+            if (block1 == Blocks.FLOWER_POT)
```
someaddons: pretty sure this would belong into the specialEqual predicate

src/main/java/com/ldtteam/structurize/util/InventoryUtils.java
```diff
@@ -0,0 +1,110 @@
...
+                    else
+                    {
+                        stack.setCount(stack.getCount() - contentCount);
+                        break;
+                    }
+                }
+            }
+        }
+
+        for (final ItemStack stack : new ArrayList<>(listToDiscount))
```
someaddons: faster: loop listToDiscount and see if any stack.getCount() > 0 -> return false. After loop return true

src/main/java/com/ldtteam/structurize/util/InventoryUtils.java
```diff
@@ -0,0 +1,110 @@
...
+        {
+            listToDiscount.remove(stack);
+        }
+
+        return listToDiscount.isEmpty();
+    }
+
+    /**
+     * Method to swap the ItemStacks from the given source {@link IItemHandler}
```
someaddons: bad javadoc

src/main/java/com/ldtteam/structurize/util/TickedWorldOperation.java
```diff
@@ -165,7 +169,47 @@ public boolean apply(final ServerWorld world)
 
         if (operation == OperationType.PLACE_STRUCTURE)
         {
-            currentPos = wrapper.placeStructure(world, storage, currentPos);
+            StructurePhasePlacementResult result;
+            switch (structurePhase)
+            {
```
someaddons, line 176: not a fan of these hardcoded int-phases :D

someaddons (commented):
Looks good overall, left some small comments

src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
Raycoms: Yeah, we could build a cached map in the Blueprint, that's true.

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Raycoms: I like to turn it on manually only. Phases that want it, turn it on manually, the other phases just don't.

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
Raycoms: It does, this list is always empty if not includeEntities

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
Raycoms: Good question, that's what I wondered about.

src/main/java/com/ldtteam/structurize/util/TickedWorldOperation.java
Raycoms: Those are only for this specific use case. Other handlers will most likely have their own phases.

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
someaddons: ye but we reset on many places, so it may get lost. when using it I'd expect the flag to not get set back to default all the time when its reseting some stuff

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
someaddons: ah ok

src/main/java/com/ldtteam/structurize/util/TickedWorldOperation.java
someaddons: still at least an enum? xD

src/main/java/com/ldtteam/structures/helpers/BlueprintIterator.java
Raycoms: If it resets, this means automatically that this iteration finished. That's the only moment it is reset.

src/main/java/com/ldtteam/structurize/util/TickedWorldOperation.java
Raycoms: An enum, for one time local use? Nah

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
Raycoms: Some doc of missing things for me =D

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
Raycoms: Would take away a lot of the power this class has. I couldn't think of any better way to handle this tbf without having a ton of duplicate code.

someaddons (approved):

src/api/java/com/ldtteam/structurize/api/util/ItemStackUtils.java
```diff
@@ -85,5 +93,46 @@ public static int getSize(final ItemStack stack)
...
+            {
+                final ItemStack stack = ((ItemFrameEntity) entity).getDisplayedItem();
+                if (!ItemStackUtils.isEmpty(stack))
+                {
+                    stack.setCount(1);
+                    request.add(stack);
+                }
+                request.add(new ItemStack(Items.ITEM_FRAME, 1));
+            }
+            else if (entity instanceof ArmorStandEntity)
```
marchermans, line 119: Would a check for IItemHandler be a good idea here?

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -35,15 +40,14 @@
      * Predicated to determine if a block is free to place.
      */
     @NotNull
-    private static final List<BiPredicate<Block, BlockState>> freeToPlaceBlocks = Arrays.asList(
+    public static final List<BiPredicate<Block, BlockState>> freeToPlaceBlocks = Arrays.asList(
```
marchermans: Formatting

marchermans (approved):
Looks good.

src/api/java/com/ldtteam/structurize/api/util/ItemStackUtils.java
Raycoms: Good question, I think if we handle itemHandler checks here that would require quite a bit of testing. This used to be a bit more generic and caused a lot of issues with other entities, that's why I removed it and made it more explicit. For the time being that is the best way to handle it.

PR #158, opened 2020-06-09
Bump optifine, add resources listener for clearing renderer cache

Nightenom:
- Add resources listener for clearing blueprint renderer cache, unfortunately it does not fix shader problems, entire mc reload required after changing optifine shaders

Review please


Raycoms:
Looks good

Raycoms:
:shipit:

PR #160, opened 2020-06-13
Prevents blocks from connecting to the substitution block

### LINKED ISSUE #153: substitution block connections
Can we change the way substitution blocks to not allow "connection" to them, like glass, fences and iron. Because when builders use them, even if there isn't anything after the build, if there isn't anything to connect the glass or fences to, it shows it connected.

image

PowerSchill:
Closes #153
Closes #
Closes #

-
-
-

Review please


PowerSchill:
image

image

Raycoms:
:shipit:

PR #173, opened 2020-07-09
Add tooltip rendering to all panes

someaddons:
Add tooltip rendering to all panes

Review please

### DIFF
--- a/src/main/java/com/ldtteam/blockout/Pane.java
+++ b/src/main/java/com/ldtteam/blockout/Pane.java
@@ -69,9 +69,6 @@ public class Pane extends AbstractGui
         super();
         id = params.getStringAttribute("id", id);
 
-        //   toolTipLines = new ArrayList<>();
-        //   toolTipLines.add(getID());
-
         @NotNull final PaneParams.SizePair parentSizePair = new PaneParams.SizePair(params.getParentWidth(), params.getParentHeight());
         PaneParams.SizePair sizePair = params.getSizePairAttribute("size", null, parentSizePair);
         if (sizePair != null)
--- a/src/main/java/com/ldtteam/blockout/Pane.java
+++ b/src/main/java/com/ldtteam/blockout/Pane.java
@@ -2,8 +2,10 @@ package com.ldtteam.blockout;
 
 import com.ldtteam.blockout.views.View;
 import com.ldtteam.blockout.views.Window;
+import com.mojang.blaze3d.systems.RenderSystem;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.AbstractGui;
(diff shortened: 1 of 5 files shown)


src/main/java/com/ldtteam/blockout/Pane.java
```diff
@@ -344,12 +352,33 @@ public final void drawLast(final int mx, final int my)
         if (visible)
         {
             drawSelfLast(mx, my);
+
+            if (isHovered && toolTipLines != null && !toolTipLines.isEmpty())
```
Raycoms: just initialize toolTipLines with an empty list so we don't have to do both checks here

src/main/java/com/ldtteam/blockout/Pane.java
```diff
@@ -805,4 +831,14 @@ public boolean onMouseDrag(final double mx, final double my, final int speed, fi
     {
         return false;
     }
+
+    /**
+     * Sets the tooltip to render on hovering this element
+     *
+     * @param lines the lines to display
+     */
+    public void setHoverToolTip(final List<String> lines)
```
Raycoms, line 842: There should probably be a way to set it via pane params too.

someaddons: nah this is easier :D

Nightenom (commented):
Probably need hooks for translation similar to what label element is using. 
Also I'm 80% sure we don't need to copy the forge class
Edit: I can finish this at Tuesday if you are not in hurry

src/main/java/com/ldtteam/blockout/Pane.java
Nightenom: Agree with Ray honestly

Raycoms: Oh, I mean we should have both in this regards.

Raycoms:
We need to copy the forge class because the forge class uses itemStacks and can't do without. We have to remove those.

src/main/java/com/ldtteam/blockout/Pane.java
someaddons: I think we only should pass essential and required information as constructor parameters, not optional ones. Else we run into a pleathora of default constructors with less parameters for each and every class extending pane

Raycoms: not in the constructor. To the paneparams object.

someaddons: should've said so to start with! lately you're very sparse with details xD

Raycoms: Marie Kondo

PR #181, opened 2020-07-19
Replace DatatypeConverter.printHexBinary; Fixes #135

### LINKED ISSUE #135: DatatypeConverter crashing the game while loading world
### Expected behavior
Getting into the world

### Actual behaviour
I am using JAVA 11 which is already supported by forges code. Not like 14+ etc.
While loading the world the game crashes with the error:

The game crashed whilst exception in server tick loop
Error: java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter

### Steps to reproduce the problem

1. Be in world selection menu
2. Load a world
3. Somehow always crashes at 8% loading it
4. ... sad

### Logs:
Debug Log: https://pastebin.com/spCXensK
Latest Log: https://pastebin.com/PQcBEWkT
Crash Log: https://pastebin.com/fhTRyk9A
P.S. great work you guys are doing on Minecolonies. JAVA 11 uses way less RAM because of the Garbage Collector thats why i think its cool to have.

Kojoley:
Replaces removed in Java 9 functionality with Apache Commons equivalent, which is also 2x faster according to https://stackoverflow.com/a/58118078/3621421.

### DIFF
--- a/src/main/java/com/ldtteam/structurize/util/StructureUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/StructureUtils.java
@@ -6,7 +6,7 @@ import net.minecraft.util.Rotation;
 import net.minecraft.util.math.Vec3d;
 import org.jetbrains.annotations.NotNull;
 
-import javax.xml.bind.DatatypeConverter;
+import org.apache.commons.codec.binary.Hex;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -45,7 +45,7 @@ public final class StructureUtils
         try
         {
             final MessageDigest md = MessageDigest.getInstance("MD5");
-            return DatatypeConverter.printHexBinary(md.digest(bytes));
+            return Hex.encodeHexString(md.digest(bytes));
         }
         catch (@NotNull final NoSuchAlgorithmException e)
         {


Raycoms:
Fine with that :shipit:

Nightenom (commented):
Since this is easy to do without any dependency I would like go that way.

Raycoms:
Yeah it's no additional dependency, it's future proof and quicker, so no need to not merge it.

Kojoley:
You are already depending on Apache Commons, what's the point?
Anyway, if you know how you want it to be done -- just do it.

Nightenom:
The point is any forge mod sits on two APIs (vanilla and Forge), thus using any dependency which we have no control over might be kind of pain in the future

Kojoley:
The change is orthogonal to Forge or MC, I do not understand that part. I did not add any additional dependency, you are already using Apache Commons, or you are asking me to remove them all? If you are in a process of removing Apache Commons you should have been said it from the beginning, so I could use Guava instead, which you are using quite extensively, or you are removing Guava too, and you are really ok with BigInt 8x slowdown?

Raycoms:
I don't see this as a problem @OrionDevelopment do you?

Nightenom:
Noone asked for removing anything, also this mod started as pull out of the main mod so it's completely a mess. As I stated we are targetting to not use any dependency which we have no control over in new PRs. Yeah, you can say apache commons or guava will probably stay here for ages but one should keep same approach for any time.
I don't care if we merge it that way or another too, I just noted this doesn't need to use dependency.
And there is tons of other ways to convert bytes[] into hex string which aren't using BigInterer

marchermans (approved):
Approved. This is a good change. Java 9 is missing the javax packages anyway, and since apache commons is a dependency of Minecraft it self this is fine.

PR #198, opened 2020-08-08
Schematic loading offthread

someaddons:
Move schematic loading offthread to no longer cause long delays on logging/startups

Review please


src/main/java/com/ldtteam/structurize/management/Structures.java
```diff
@@ -88,13 +93,18 @@
...
      * file extension for the schematics
      */
     @NotNull
-    private static final Map<String, String> fileMap = new HashMap<>();
+    private static final Map<String, String> fileMap = new ConcurrentHashMap<>();
+
+    /**
+     * Offthread schematic reading
+     */
+    private static ThreadPoolExecutor executor = (ThreadPoolExecutor) Executors.newCachedThreadPool();
```
Nightenom: Can we use vanilla thread executor?

src/main/java/com/ldtteam/structurize/management/Structures.java
```diff
@@ -119,7 +129,7 @@ private Structures()
      */
     public static void init()
     {
-        loadStyleMaps();
+        executor.submit(() -> loadStyleMaps());
```
Nightenom: Add boolean and make it synchronous by default (because of reload method below).

Nightenom (changes requested):
I'm ok with this for client, for server this definitely needs joining before players can connect. (For client it would be cool to join that before player tries to connect or before main menu screen)
In which load phase this happens?

malte0811:
I have some profiling data for this from logging into ForgeCraft, nearly all of the time is in `StructureLoadingUtils#getStream`:
image

About 40% that time will probably be fixed by https://github.com/MinecraftForge/MinecraftForge/pull/7256. The rest is mostly spent in [this method](https://github.com/MinecraftForge/MinecraftForge/blob/1.16.x/src/fmllauncher/java/net/minecraftforge/fml/loading/LoadingModList.java#L121-L136), so that part of the time grows linearly with mod count. I don't know whether using resource/data pack methods is possible here, they do currently have the same issue but will be fixed with https://github.com/MinecraftForge/MinecraftForge/pull/7228.

 You're currently calling `getStream` twice on (as far as I can tell) the same file, so you can get rid of half the time easily by just storing the file contents in memory.

someaddons:
ye this is where it loads schematics from file, and as we have quite a lot ~3400+ it takes a while

malte0811:
It probably wouldn't "take a while" without the Forge issues (and only opening the files once). The first one I mentioned should effectively get rid of those 40%, and from my experience in IE using a constant-time method for getting resources can give a factor 10 improvement in larger packs (see https://github.com/MinecraftForge/MinecraftForge/issues/7227).

Raycoms:
@Nightenom what is the worst that can happen here? Schematic missing when opening build tool?

Raycoms:
When she restarts does it fix it?

malte0811:
[This patch](https://gist.github.com/malte0811/feec55ca151f13f99618654a2b258a09) gets structure loading time for MineColonies down to <3 seconds when joining FC (from 30-40 seconds). I applied the patch to cd556d705b4d6dec310321e5765e979efa1edf4d, the last 1.16.1 commit. I'm not sure whether you'll want that code (it does a lot of things manually that happen automatically otherwise), if you do you can either just apply the patch directly or I can create a PR. The latter would probably be a good idea since the code style is a wild mix of IEs style, Structurize's style and my own style. The `ThreadLocal`s might be unnecessary, I don't know whether the code is called on multiple threads and `ThreadLocal` is cheap enough for me to be happy using it for testing.

Nightenom:
@malte0811 looks hell promising, it's kinda funny how long does it take for forge to discover ModFile. We will take your idea and put it into next PR, will mention you in commits. Thanks for your time and work you have put into this

Nightenom:
Closing this since there is better single thread solution

PR #204, opened 2020-08-15
Port to 1.16.2

marchermans:
This ports structurize to 1.16.2 including the changes proposed by @Nightenom from #203


build.gradle
```diff
@@ -25,9 +24,6 @@ repositories {
         name 'LDTTeam - Modding'
         url 'https://ldtteam.jfrog.io/ldtteam/modding/'
     }
-    flatDir {
```
Nightenom, line 28: Please don't remove libs, I use them quite often

src/main/java/com/ldtteam/structures/client/BlueprintBlockAccess.java
```diff
@@ -10,6 +10,9 @@
 import com.ldtteam.structures.blueprints.v1.Blueprint;
```
Nightenom, line 10: My generic approach in this class was supply null whereever possible and then test and fix crashes according our needs.
Will clean it up myself in my next PR.

Nightenom (approved):
Nice, just what is `LanguageMap.getInstance().func_241870_a(renderText)`?
Note: we used IFormattableTextComponent over ITextProperties because it had append methods (probably it still has), classes ITextComponent and ITextProperties were the only super classes and there weren't any other subclasses of them (not sure if that changed, don't know what is IReorderingProcessor). Thus I would like to keep using this interface instead.

build.gradle
marchermans: You can use libs via mavenLocal.
That also allows you to use fg.deobf.

Never use a flatDir since FG3 does not support that officially and also not fully

marchermans:
@Nightenom There is now a type split between `ITextComponent`-types and the type returned by the split function of fontRenderer. Which is now a different type.
That LanguageMap call is needed to convert ITextComponent to that special component.

build.gradle
Nightenom: Welp, I keep using flatDir since 1.13 without single issue
I don't have set up mavenLocal but guess I should learn something new...

PR #205, opened 2020-08-18
Faster schematic discover from jar files

Nightenom:
- reworked loading from jars based on [patch](https://gist.github.com/malte0811/feec55ca151f13f99618654a2b258a09) by @malte0811
- removed unnecessary duplicated getStream() calls

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/management/Structures.java
+++ b/src/main/java/com/ldtteam/structurize/management/Structures.java
@@ -156,7 +156,7 @@ public final class Structures
 
         if (md5Map.size() == 0)
         {
-            Log.getLogger().error("Error loading StructureProxy directory. Things will break!");
+            Log.getLogger().warn("No file found during schematic discover. Things may break!");
         }
     }
 
@@ -215,23 +215,18 @@ public final class Structures
             while (it.hasNext())
             {
                 final Path path = it.next();
-                Log.getLogger().warn("Orig: {}", path.toString());
                 if (path.toString().endsWith(SCHEMATIC_EXTENSION_NEW))
                 {
                     String relativePath = basePath.relativize(path).toString();
-                    Log.getLogger().warn("Prefix: {}", relativePath);
                     relativePath = relativePath.substring(0, relativePath.length() - SCHEMATIC_EXTENSION_NEW.length());
-                    Log.getLogger().warn("Interfix 1: {}", relativePath);
                     if (!SCHEMATICS_SEPARATOR.equals(path.getFileSystem().getSeparator()))
                     {
                         relativePath = relativePath.replace(path.getFileSystem().getSeparator(), SCHEMATICS_SEPARATOR);
                     }
-                    Log.getLogger().warn("Interfix 2: {}", relativePath);
                     if (relativePath.startsWith(SCHEMATICS_SEPARATOR))
                     {
                         relativePath = relativePath.substring(1);
                     }
-                    Log.getLogger().warn("Postfix: {}", relativePath);
 
                     try
                     {
@@ -239,7 +234,6 @@ public final class Structures
                         fileMap.put(structureName.toString(), SCHEMATIC_EXTENSION_NEW);
                         final byte[] structureFileBytes = StructureLoadingUtils.getByteArray(relativePath);
                         final String md5 = StructureUtils.calculateMD5(structureFileBytes);
-                        Log.getLogger().warn("SN: {}, MD5: {}", structureName.toString(), Objects.toString(md5));
                         if (md5 == null)
                         {
                             fileMap.remove(structureName.toString());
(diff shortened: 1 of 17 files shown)


src/main/java/com/ldtteam/structurize/management/Structures.java
```diff
@@ -170,95 +158,6 @@ private static void loadStyleMaps()
...
-                final URL url = Structures.class.getResource(SCHEMATICS_ASSET_PATH + origin);
-                uri = url.toURI();
-            }
-            catch (@NotNull final URISyntaxException e)
-            {
-                Log.getLogger().error("loadStyleMaps : ", e);
-                return;
-            }
-
-            if ("jar".equals(uri.getScheme()))
```
Raycoms, line 193: Don't we need this anymore? I know that a lot of this was compatibility for different OS

Raycoms: + Different origin for dev env and for real game

Nightenom:
@malte0811 your review would be appreciated

src/main/java/com/ldtteam/structurize/util/StructureLoadingUtils.java
```diff
@@ -28,7 +32,16 @@
     /**
      * The list of origin folders.
      */
-    public static List<String> originFolders = new ArrayList<>();
+    private static final List<String> originFolders = new ArrayList<>();
+    /**
+     * The list of origin mods.
+     */
+    private static final Map<String, ModFileInfo> originMods = new HashMap<>();
```
malte0811, line 39: Theoretically you can get something like mod A adding resources in mod B's namespace (so you'd need to check all mods to know a resource doesn't exist), but I guess you are intentionally ignoring that possibility?

malte0811 (dismissed):
Looks good to me.

src/main/java/com/ldtteam/structurize/util/StructureLoadingUtils.java
Nightenom: Yep, I wondered why you were putting the result from searching through all mods into the modid<->modfile map. You can now completely override mod entry (in this map) by addOriginMod() method. I would like to use resource packs instead (and remove this mess) but I was told that there is some problem with that...

src/main/java/com/ldtteam/structurize/management/Structures.java
marchermans: This is still needed if he does not handle it differently elsewhere.

src/main/java/com/ldtteam/structurize/management/Structures.java
```diff
@@ -297,11 +198,12 @@ private static void checkDirectory(@NotNull final File directory)
...
      * and add them in the md5Map
      *
-     * @param basePath the base path.
+     * @param base     the base path.
      * @param prefix   either schematics, scans, cache
      */
-    private static void loadSchematicsForPrefix(@NotNull final Path basePath, @NotNull final String prefix)
+    private static void loadSchematicsForPrefix(@NotNull final Path base, @NotNull final String prefix)
     {
+        final Path basePath = base.toAbsolutePath();
```
marchermans, line 206: Not going to work in an obfuscated environment.
This needs to take the scheme into account -> Which means you need to use at least a URI.

marchermans (changes requested):
Personally: This is not the way to go!
If this is reworked, this should be using datapacks and IFutureReloadListeners to load everything asynchronously and then process then on the main thread if even needed (if you implement the backing registry with concurrency in mind then you can do it all asynchronously).

That said, the new code will not work once packed into a jar. Since Paths do not care for the fact that it is in a jar or not.

src/main/java/com/ldtteam/structurize/management/Structures.java
Nightenom: It works, it's tested well

Nightenom:
Orion: load datapack from dedi server (to client) or load resource pack on dedi server, if you can manage to get one this working then I would like to use it too. Since either of these seems to be not feasible to me
btw this is not a rework, it's just two fixes: one on our end and the second is to mitigate forge behaviour by handling it completely on our end

src/main/java/com/ldtteam/structurize/management/Structures.java
Nightenom: Btw exactly this was added to fix handling of the damned ZipPath (which is used when loading from jar) which was not working well with relativize

Nightenom:
> That said, the new code will not work once packed into a jar. Since Paths do not care for the fact that it is in a jar or not.

Nope, wrong, afaik cpw added handling for this with the new mod launcher. If you look carefully this fix is using modFile locator which is also used for mods.toml/accesstransformer.cfg discover and few others

marchermans:
Oh okey
If that is added then I am fine,
did not know that.

Learned something today.

PR #210, opened 2020-09-04
Trigger entity success

Raycoms:
---- API CHANGE ----
Does not work with older Minecolonies versions
----------------------------

- Adds an entity success trigger which the builder needs
- 
-

Review please


Nightenom (approved):

someaddons (approved):

marchermans:
Why was this assigned to me? Is there backwards compatibility issues?

Nightenom:
It adds method to api iface which may result in crash when implementation is missing

marchermans:
So? The only reason this is a problem, is when a Minecolonies version is supplied that is older then this Structurize version. Limit the Structurize version in MC to require at least this version. There is no backwards compatibility issue here. And I think we do not want to limit us to forward compatibility as well @Raycoms ?

PR #233, opened 2020-11-22
fix #6118

Raycoms:
- Fixes https://github.com/ldtteam/minecolonies/issues/6118
-
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
+++ b/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
@@ -119,7 +119,7 @@ public final class PlacementHandlers
             itemList.removeIf(ItemStackUtils::isEmpty);
             if (!world.getBlockState(pos.down()).getMaterial().isSolid())
             {
-                itemList.add(BlockUtils.getItemStackFromBlockState(BlockUtils.getSubstitutionBlockAtWorld(world, pos)));
+                itemList.addAll(getRequiredItemsForState(world, pos, BlockUtils.getSubstitutionBlockAtWorld(world, pos), tileEntityData, complete));
             }
             return itemList;
         }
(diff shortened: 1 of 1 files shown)


marchermans (approved):

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -714,6 +714,27 @@ public static void handleTileEntityPlacement(
...
+     * @param complete if complete.
+     * @return the required items.
+     */
+    public static List<ItemStack> getRequiredItemsForState(final World world, final BlockPos pos, final BlockState state, final CompoundNBT data, final boolean complete)
+    {
+        for (final IPlacementHandler placementHandler : PlacementHandlers.handlers)
+        {
+            if (placementHandler.canHandle(world, pos, state))
+            {
+                return placementHandler.getRequiredItems(world, pos, state, data, complete);
```
Nightenom, line 732: This can cause an infinite loop as you are calling this from getRequiredItems

Nightenom (changes requested):

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Raycoms: Theoretically yes, practically this won't happen.

Nightenom: Funny not funny, go merge otherwise

ab9rf: > Theoretically yes, practically this won't happen.

It's happening to me right now. Thought you might want to know.

Raycoms: Please make a github issue with the logs, this is very old.

ab9rf: we are getting a stack overflow bouncing between `PlacementHandlers.java:258` and `PlacementHandlers.java:1071` but i have yet to nail down a minimal combination of mods that elicits the behavior. we know at this point that adding architectury to a set of mods that works causes it to stop working, but adding architectury on its own doesn't cause the crash. (unfortunately my son forgot to snapshot the list of mods that works without architectury but breaks with architectury)

i'm not going to open an issue until we have that solution because my past interactions with your team tell me that if i don't put in that _entire_ effort in advance y'all wont even look at the problem, and while i used to be a java developer that was over a decade ago and i don't have a working java build or debug environment here so i can't just use an instrumented build to find out what block(s) is causing the recursion even though that's what i'd do if i had that option. unfortunately i'm a C++ developer these days and the toolsets i have readily at hand are not much use in diagnosing issues with java code

Raycoms: If you make a github issue, with a lot that shows the list of mods & you're on the most up to date version, we will definitely look at it. We might still need to find out which mod it is. But we can already rule out some things then.

PR #234, opened 2020-12-02
Fluid Placeholder Block

ShadowProtocol:
- Adds a new type of substitution/placeholder block: the Fluid Placeholder
    - Ensures that the destination block is solid or a fluid source
    - If the destination block is not a fluid source already, the default fluid type for the dimension is used (i.e. lava in nether)
    - renders in the preview as the default fluid type for that dimension

some example use cases:
- almost all underwater structures should use this over a regular Placeholder block (this is part of the setup for underwater styles)
- around the edge of ship schematics so that they don't take chunks out of the shore when placed close to land.
- any decorations that work with the respective dimension fluid, for example a fountain that is a lava fountain in the nether

Requires a consecutive PR to minecolonies but works already when stand-alone.

Review please


src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -63,6 +97,44 @@ private PlacementHandlers()
          */
     }
 
+    public static class FluidSubstitutionPlacementHandler implements IPlacementHandler
+    {
+        BlockState existing;
```
Raycoms: I don't think this existing var here is a good idea at all.

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -63,6 +97,44 @@ private PlacementHandlers()
...
+        @Override
+        public ActionProcessingResult handle(
+          @NotNull World world,
+          @NotNull BlockPos pos,
+          @NotNull BlockState blockState,
+          @Nullable CompoundNBT tileEntityData,
+          boolean complete,
+          BlockPos centerPos)
+        {
+            if (existing.hasProperty(BlockStateProperties.WATERLOGGED))
```
Raycoms: missing formatting

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -395,6 +410,18 @@ public static void removeFluid(World world, BlockPos pos)
...
     }
 
+    /**
+     * A simple check to fetch the default fluid block for this dimension
+     * @param world the world of the dimension
+     * @return the default blockstate for the default fluid
+     */
+    public static BlockState getFluidForDimension(World world)
+    {
+        return world.getDimensionKey().equals(World.THE_NETHER)
```
Raycoms: This is only for The nether and not for netherlike dimensions though. Since it's not taking the DImenstionType

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Nightenom: agree, never do that

ShadowProtocol: So I've been considering this throughout the day. I was aware it is an unorthadox solution, and originally I did it just so I could get testing. But I'm not really sure what the best alternative really is. I think this is a unique case where instead of removing something or ignoring it, the fluid block needs to take in the block that was already there so it can waterlog it.

I guess I could just inline all this into the main ``handleBlockPlacement``, but I was preferring to avoid putting exceptions in engine-like code (even though you already have some for the solid placeholder anyway).

Raycoms: can't we query the existing block in the handle?

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Nightenom: now you can only waterlogged water (if should contain water check)
and you really should for whatever reason query `existing` here from world and pos
and also iirc `complete `should be taken into account here too - it should place placeholder itself when pasting instead of placing (look at other placeholders)

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Nightenom: the `canHandle` check should be just `instanceof BlockFluidSubstitution`

ShadowProtocol: oh I'm completely blind I did not see that ``worldPos`` was passed in to the handler. That could have saved me a bit of effort. I must've mixed it up for being a local position or something ... idk.

Raycoms: =D

ShadowProtocol: OK I'm not as blind as I seem; I remember the reason I originally did it. The block is removed before the new one is handled, so querying by worldPos doesn't work because there's no longer a block there to waterlog. I'm thinking of just handling this as an exception inline, where I have access to the ``worldState`` variable.

Raycoms: What we can do, might be a bit more effort, is also create in the handlers a removalHandler, and call that instead of the fixed onRemove in code, then all the existing handlers just use the default onRemoval in the interface, and the substitution blocks handle it themselves?

ShadowProtocol: That certainly does seem much more extensible. I like it. I'll have a go implementing that when my schedule frees up later this week.

Nightenom: @ ShadowProtocol you should be able to adapt the calling of handler methods to include the old block, (try to use defaulting methods as Ray suggested)

Raycoms:
This looks good now :shipit:

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -63,6 +99,64 @@ private PlacementHandlers()
...
+        @Override
+        public boolean canHandle(@NotNull World world, @NotNull BlockPos pos, @NotNull BlockState blockState)
+        {
+            return blockState.getBlock() instanceof BlockFluidSubstitution;
+        }
+
+        @Override
+        public List<ItemStack> getRequiredItems(@NotNull World world, @NotNull BlockPos pos, @NotNull BlockState blockState, @Nullable CompoundNBT tileEntityData, boolean complete)
+        {
+            return new ArrayList<>();
```
Nightenom: not sure if this should return bucket with fluid

Nightenom (dismissed):
One minor thing about resource handling, otherwise ok

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Raycoms: oh, at least for lava it should

Nightenom (approved):

PR #236, opened 2020-12-06
Fix #235

### LINKED ISSUE #235: [1.16.4] CTD When Replacing Oak->Spruce Door
### Prerequisites 

- [x] I am running the latest **alpha** version of Structurize for my Minecraft version.
- [x] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure my issue is not covered there.
- [x] I made sure that **this issue is not a duplicate of any existing issue.**

### Context 

- Minecraft Version: 1.16.4
- Structurize Version: structurize-0.13.100-RELEASE-universal.jar
- MineColonies Version (if related bug): minecolonies-0.13.486-ALPHA-universal.jar
- Related Mods and their Versions: All the Mods 6 1.3.1

## Expected behavior
Replaces door without CTD

## Actual behavior
Crashes to desktop when replacing a door. Door was originally oak and attempted to replace to spruce.

## Steps to reproduce the issue

I was editing a pasted-in schematic (wooden enchanter's tower, level 1). Attempted to use structurize to replace the oak door with a spruce door and got a CTD.

### Logs

- crashlog: https://gist.github.com/Sandriell/c011f21febc33acc3cfd124e6e41ebad

### Notes

---
## Viewers

Nightenom:
Closes #235

- updated gradle and Forge
- partial fix for placing of BlockItems, a full fix is literally impossible:
new behaviour for BlockItems is:
  - are the target and current (in world) block the same class? just carry over blockstate properties
  - do the target and current have superClass? carry common properties
  - are they completely different? try to create a blockstate, this is the old way and it's a lottery guess for now (because of facings etc.)

Review please


Raycoms:
Looks good :shipit:

Nightenom:
Fixed tons of handling, incl. air being counted twice in the scan tool, implemented correct bucket handlers
but not sure if anyone uses IFluidStack and those things for fluids added by Forge since vanilla is able to handle too imho

Raycoms:
:shipit: waiting for @OrionDevelopment

gradle.properties
```diff
@@ -7,7 +7,7 @@ base_group=com.ldtteam
 base_mod_id=structurize
 
 minecraft_version=1.16.4
-forge_version=35.0.1
+forge_version=35.1.7
```
marchermans: This is not the latest.
When we update, we should probably either go to the lastest recommended, or if that is not selected (like here) then we should go to the latest available version.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,4 @@
-#Sat Jan 11 07:01:08 BRT 2020
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.3-bin.zip
```
marchermans, line 1: Why?
I am pretty sure this is not needed.

src/main/java/com/ldtteam/structurize/client/gui/WindowReplaceBlock.java
```diff
@@ -165,10 +174,25 @@ public void onButtonClicked(@NotNull final Button button)
         if (button.getID().equals(BUTTON_DONE))
         {
             final ItemStack to = findPaneOfTypeByID("resourceIconTo", ItemIcon.class).getItem();
-            if (!ItemStackUtils.isEmpty(to))
+            if (!ItemStackUtils.isEmpty(to) || to.getItem() instanceof AirItem)
```
marchermans, line 177: This makes no sense.
Why are we checking here for Air?
Please document this better.

src/main/java/com/ldtteam/structurize/client/gui/WindowReplaceBlock.java
```diff
@@ -124,38 +140,31 @@ public void onOpened()
...
         findPaneOfTypeByID("resourceNameTo", Label.class).setLabelText((IFormattableTextComponent) new ItemStack(Blocks.AIR).getDisplayName());
         updateResources();
+        updateResourceList();
     }
 
     private void updateResources()
     {
         allItems.clear();
         allItems.addAll(ImmutableList.copyOf(StreamSupport.stream(Spliterators.spliteratorUnknownSize(ForgeRegistries.ITEMS.iterator(), Spliterator.ORDERED), false)
+            .filter(item -> item instanceof AirItem || item instanceof BlockItem || (item instanceof BucketItem && ((BucketItem) item).getFluid() != Fluids.EMPTY))
```
marchermans, line 150: Again this feels really like a code smell with the Air check.....

marchermans (changes requested):
Okey soooooo....
I am strugling with this PR.

Primarily because i see no usecase.
Could you document this better?

Screenshots of things that this fixes would be a start.

I am pointing this out, since:
1) You are not fixing the problem fully
2) The way you are fixing it, goes against everything vanilla MC does when it copies in its own schematics. So i am not sure what the problem is.

Also see comments on the version updates.

gradle.properties
Nightenom: pr was started before 35.1.10, sure can update

gradle/wrapper/gradle-wrapper.properties
Nightenom: like im having tons of versions of gradle just because every project use different version and different classifier
btw this is mdk recommended version

src/main/java/com/ldtteam/structurize/client/gui/WindowReplaceBlock.java
Nightenom: air itemstacks are considered empty itemstacks
yep, we have remove button but also we have air block in resource listing, this is just quality of life change

src/main/java/com/ldtteam/structurize/client/gui/WindowReplaceBlock.java
Nightenom: AirItem is **not** BlockItem, ask mojang why..

Nightenom:
> Primarily because i see no usecase.
> Could you document this better?

This pr is nothing but wide fix to #235, it includes tons of qol changes and things around that were old or not as full-covering as they could be

> 1. You are not fixing the problem fully

it's not possible because block replacement is always gonna be hard in terms of which properties should be carried and which not
this would require extensive gui with nbt editor which I'm not going to implement now 

> 2. The way you are fixing it, goes against everything vanilla MC does when it copies in its own schematics. So i am not sure what the problem is.

this function has nothing to do with scanning/placing (it has just wrong name), it's just a replace function and it lacks user input as said above, so it tries the best even for blocks with many properties

marchermans (approved):

PR #247, opened 2020-12-18
Blockout/rework text elements

Nightenom:
Sup PR of fullscreen-guis to keep it sane

- rework AbstractTextElement to merge (unify) all label and text rendering and remove code dupes, removed Label class and backed label element with text class, tried to reduce needed XML changes to minimum
- I know this is breaking change in code for many guis, will fix it myself on mcol side so you don't blame me for it
- I also know the nearly all labels don't have their sizes set correctly thus rendering not rendering fully, will try to fix it myself too
- fixed gui sizes
- disabled vanilla crosshair in our guis

Review please


src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -37,7 +37,7 @@
         register("text", Text.class);
         register("button", ButtonVanilla.class);
         register("buttonimage", ButtonImage.class);
-        register("label", Label.class);
+        register("label", Text.class); // we don't want to deal with xml changes
```
Raycoms: maybe add a "todo" remove in the future or so

src/main/resources/assets/structurize/gui/windowstructurenameentry.xml
```diff
@@ -1,6 +1,6 @@
 <window xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../../../java/com/ldtteam/blockout/blockOut.xsd"
         pause="true" lightbox="true">
-    <label size="100% 11" pos="0 100" textalign="TOP_MIDDLE" color="white" label="$(com.ldtteam.structurize.gui.structure.edit.title)"/>
+    <label size="100% 11" pos="0 100" textalign="TOP_MIDDLE" color="white" label="Structure path:"/>
```
Raycoms: Why shouldn't this be a translation string?

Nightenom: I was dead translation, I was just testing things, need to add it into lang first

src/main/java/com/ldtteam/blockout/Loader.java
Nightenom: will probably nuke it too, cuz with all these three PRs and upcoming Shadow's changes it's gonna be a quite big breaking change already

src/main/java/com/ldtteam/blockout/controls/AbstractTextElement.java
```diff
@@ -1,101 +1,392 @@
...
+
+        ms.pop();
+    }
+
+    public Alignment getTextAlignment()
+    {
+        return textAlignment;
+    }
+
+    public void setTextAlignment(final Alignment textAlignment)
```
Raycoms, line 252: missing javadoc

Raycoms: a bunch of methods in this class

Nightenom: Do we really need jdoc for set/get methods?

Nightenom: @ShadowProtocol what do you think of this?

Nightenom: Same probably applies to the text builder

ShadowProtocol: I just imagine it as if it is in an auto-genned doc page. would someone understand that this sets the text alignment based off of the method title alone? If the answer is yes then it doesn't need a jdoc IMO. You just need to be careful that you aren't doing other things in the setter that won't be described by the title alone. 
But if you're just going to chuck ``/** Sets the text alignment */`` above it then that is not worth anyone's energy scrolling past it.

PR #249, opened 2020-12-24
Gui hooks for entities and tileentities

Nightenom:
- adds ability to render gui above entities and tileentities in the same way as vanilla entity name is rendered
- gui scale is 100 px = 1 block
- naming might be weird since my brain is having a weird week, feel free to suggest changes

Review please

### DIFF
--- a/src/main/java/com/ldtteam/blockout/Loader.java
+++ b/src/main/java/com/ldtteam/blockout/Loader.java
@@ -27,6 +27,8 @@ import javax.xml.parsers.ParserConfigurationException;
  */
 public final class Loader
 {
+    private static final DocumentBuilderFactory DOCUMENT_BUILDER_FACTORY = DocumentBuilderFactory.newInstance();
+
     private static final Map<String, Constructor<? extends Pane>> paneConstructorMap = new HashMap<>();
     static
     {
@@ -180,8 +182,7 @@ public final class Loader
     {
         try
         {
-            final DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
-            final DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+            final DocumentBuilder dBuilder = DOCUMENT_BUILDER_FACTORY.newDocumentBuilder();
             final Document doc = dBuilder.parse(input);
             input.getByteStream().close();
 
--- a/src/main/java/com/ldtteam/blockout/hooks/HookManager.java
+++ b/src/main/java/com/ldtteam/blockout/hooks/HookManager.java
@@ -3,19 +3,26 @@ package com.ldtteam.blockout.hooks;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.BiPredicate;
 import java.util.function.Function;
 import com.ldtteam.blockout.hooks.TriggerMechanism.Type;
-import com.ldtteam.structurize.api.util.constant.Constants;
 import com.mojang.blaze3d.matrix.MatrixStack;
 import net.minecraft.client.Minecraft;
 import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.registries.IForgeRegistryEntry;
 
+/**
+ * Core class for managing and handling gui hooks
+ * 
+ * @param <T> instance of U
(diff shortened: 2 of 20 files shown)


src/main/java/com/ldtteam/blockout/hooks/HookScreen.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public void init()
+    {
+        // noop
+    }
+
+    @Override
+    public void tick()
```
Raycoms, line 64: does this have to tick?

src/main/java/com/ldtteam/blockout/hooks/HookScreen.java
```diff
@@ -0,0 +1,88 @@
...
+
+import com.ldtteam.blockout.BOScreen;
+import com.ldtteam.blockout.hooks.TriggerMechanism.Type;
+import com.ldtteam.blockout.views.ScrollingList;
+import com.mojang.blaze3d.matrix.MatrixStack;
+
+/**
+ * Screen wrapper.
+ */
+public class HookScreen extends BOScreen
```
Raycoms, line 11: this class is missing @Overrides

src/main/java/com/ldtteam/blockout/hooks/HookManager.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.Set;
+import java.util.function.Function;
+import com.ldtteam.blockout.hooks.TriggerMechanism.Type;
+import com.ldtteam.structurize.api.util.constant.Constants;
+import com.mojang.blaze3d.matrix.MatrixStack;
+import net.minecraft.client.Minecraft;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+public abstract class HookManager<T, U extends IForgeRegistryEntry<U>, K>
```
Raycoms, line 26: this should also have an interface with javadoc for the methods

src/main/java/com/ldtteam/blockout/hooks/HookScreen.java
Nightenom: nah, those methods are new methods, if you mean some concrete methods then name them (I think I have covered all of them)

src/main/java/com/ldtteam/blockout/hooks/HookManager.java
Nightenom: you mean something like IHookManager? why? this is not part of visible API imho (basically, you only want to extend it)

src/main/java/com/ldtteam/blockout/hooks/HookScreen.java
Nightenom: the gui can have one scrolling list, so probably yes

src/main/java/com/ldtteam/blockout/hooks/HookManager.java
Raycoms: Mostly to have a nice place for all the docu

src/main/java/com/ldtteam/blockout/hooks/HookScreen.java
Raycoms: ah okay, those are missing some javadoc then =D, I thought the render method at least

src/main/java/com/ldtteam/blockout/hooks/HookManager.java
Nightenom: Like there is nothing to document, can document registerInternal but that's all, tick and render are quite obvious

src/main/java/com/ldtteam/blockout/hooks/HookScreen.java
Nightenom: kk

src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -180,8 +182,7 @@ private static void createFromXML(final InputSource input, final View parent)
     {
         try
         {
-            final DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
-            final DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+            final DocumentBuilder dBuilder = DOCUMENT_BUILDER_FACTORY.newDocumentBuilder();
```
Nightenom, line 185: This should be reusable too, but I'm not sure so let's rather be safe

marchermans (approved):

PR #258, opened 2021-01-08
Blockout properties

ShadowProtocol:
- Uses resource locations for elements, and removes reflection
- Adds a pane parameters cache (shaves an average of at least 6 ms off a window over multiple samples)
- Places parsers in their own class for extra flexibility and better reuse. Simplifies attribute getters.
- Improves shorthand and text attribute getters.parsers
- Adds a toggle button (can be used a lot in mcol) for simplified state switching
- Some other miscellaneous readability adjustments

And yes, is missing some jdoc, and yes is a massive diff, sorry about that. Hopefully an improvement, but less impactful than the intent I started out with.
It also probably needs more extensive testing

Review please - go on, tear it to shreds!


src/main/java/com/ldtteam/blockout/Color.java
```diff
@@ -54,21 +57,38 @@ public static int getByName(final String name, final int def)
         return i != null ? i : def;
     }
 
+    /**
+     * Get a color integer from its name.
+     *
+     * @param name name of the color.
+     * @return the color as an integer.
+     */
+    public static int getByName(final String name)
```
Nightenom: func dupe? the above is the same and this will npe if entry is not in the map

src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -29,92 +30,90 @@
...
-        register("treeview", TreeView.class);
+        register("view", View::new);
+        register("group", Group::new);
+        register("scrollgroup", ScrollingGroup::new);
+        register("list", ScrollingList::new);
+        register("text", Text::new);
+        register("button", Button::new);
+        register("buttonimage", Button::new);
+        register("toggle", ToggleButton::new);
+        register("label", Text::new);
```
Nightenom: missing todo comment

src/main/java/com/ldtteam/blockout/PaneParams.java
```diff
@@ -140,22 +124,21 @@ private static String localize(final String str)
...
      * Get the String attribute from the name and definition.
      *
-     * @param name the name.
-     * @param def  the definition.
+     * @param name      the name.
+     * @param fallback  the default value.
      * @return the String.
      */
-    public String getStringAttribute(final String name, final String def)
+    public String string(final String name, final String fallback)
```
Nightenom: def seems more used to me rather than fallback

src/main/java/com/ldtteam/blockout/PaneParams.java
```diff
@@ -140,22 +124,21 @@ private static String localize(final String str)
      * @param name the name to search.
      * @return the attribute.
      */
-    public String getStringAttribute(final String name)
+    public String string(final String name)
     {
-        return getStringAttribute(name, "");
+        return string(name, null);
```
Nightenom: I assume you have checked this does not cause npe

src/main/java/com/ldtteam/blockout/properties/Parsers.java
```diff
@@ -0,0 +1,168 @@
...
+import static com.ldtteam.blockout.Log.getLogger;
+
+public final class Parsers
+{
+    private Parsers() { /* prevent construction */ }
+
+    /**
+     * Just a convenient way to put the Function lambda
+     */
+    public interface Any<T>
```
Nightenom: Function<String,T> might be more readable then introducing new iface
missing @ FunctionalInterface

src/main/java/com/ldtteam/blockout/PaneParams.java
```diff
@@ -93,45 +84,38 @@ public String getText()
...
-
-            index = s.indexOf("$(", index + replacement.length());
         }
 
-        return s;
+        final Node attr = getAttribute(name);
+        if (attr != null) result = parser.apply(attr.getNodeValue());
+
+        propertyCache.put(name, result);
+        return result != null ? result : fallback;
```
Nightenom: I think that null handling should be faster when using cache

src/main/java/com/ldtteam/blockout/PaneParams.java
```diff
@@ -212,243 +173,105 @@ public boolean hasAttribute(final String name)
...
      * Get the integer attribute from name and definition.
      *
-     * @param name the name.
-     * @param def  the definition.
+     * @param name     the name.
+     * @param fallback the default value.
      * @return the int.
      */
-    public int getIntAttribute(final String name, final int def)
+    public int numeral(final String name, final int fallback)
```
Nightenom: Please retain get int /float etc.

src/main/java/com/ldtteam/blockout/controls/Button.java
```diff
@@ -3,46 +3,134 @@
...
+
+    protected static final TextureRepeatable VANILLA_TEXTURE = new TextureRepeatable(
+      TEXTURE,
+      0, ENABLED_TEXTURE_V,
+      DEFAULT_BUTTON_WIDTH, DEFAULT_BUTTON_HEIGHT,
+      TEXTURE_INNER_U_OFFSET, TEXTURE_INNER_V_OFFSET,
+      TEXTURE_INNER_U_WIDTH, TEXTURE_INNER_V_HEIGHT
+    );
+
+    public    ButtonText        text;
```
Nightenom: Dont use subelements unless needed

src/main/java/com/ldtteam/blockout/properties/Tooltip.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public void draw(final MatrixStack ms, final Pane pane, final double mx, final double my)
+    {
+        Window window = pane.getWindow();
+        int width = renderedWidth + 8;
+        int height = renderedHeight + 8;
+
+        if (!preparedText.isEmpty())
+        {
+            x = (int) mx + CURSOR_BOX_SIZE - 4;
```
Nightenom: there seems to be variable misinterpretation against my version
x was a hook to pane position (since the element was moving according to the mouse pos)
textOffsetX was fixed value offsetting the text from element start

also, the tooltip is rather a hover element than pane extension - imho it should sit in the element tree

is this in drawLast pass?
try to avoid z levelling cuz there is upcoming forge PR from gigaherz which will adjust this a bit and I would like to a PR after that to nuke out the drawLast thingy

src/main/java/com/ldtteam/blockout/views/DropDownList.java
```diff
@@ -167,7 +163,7 @@ private void onButtonClickedFromList(@NotNull final Button buttonIn)
         final Text idLabel = buttonIn.getParent().findPaneOfTypeByID("id", Text.class);
         if (idLabel != null)
         {
-            final int index = Integer.parseInt(idLabel.getTextAsString());
+            final int index = Integer.parseInt(idLabel.text.get().getString());
```
Nightenom: this is kinda not intuitive to get the first row of text

Nightenom (commented):
_I'm a kinda negative person so feel free to fight back_

Personally, I would use get names for params (like getString, getBoolean)

I kinda unhappy with using subelements rather than extending, further discussion needed
We can use interfaces however
edit: we can't sometimes since abstract gui, that's why I would stick to extending
I know that there is a problem with button needing both of the extensions, I will replace the image extension with image processor (designing it right now, would rather use interface for that)
Overall, I wouldn't make text things subelement

Can you make a small image of the toggle button so we can see how it works?

However, I'm happy with the parsing changes, except the method names

ShadowProtocol:
I can see there are definitely a couple of things I need to look into. I'll get around to addressing some of that further when I get the opportunity (it's late rn). And sorry I type **a lot**:

As far as I'm aware, using group classes is an excellent design practice when there threatens to be two or more superclasses that require alternative properties (thus cannot be interfaces). It increases flexibility, where different groups can be used in different combinations and can behave completely differently while cooperating. It increases clarity, because there are less supers of supers of supers to understand which one is affecting the element at what time and which variable is defined where. It increases maintainability, because changes can be made to groups with a reduced impact on the pane structure. It also increases code safety if done well (which I don't claim to have done), because construction can be more tightly controlled for the groups purposes alone.
I know there are disadvantages though, but I still think this is a worthwhile choice,

I've seen noun-getters used before where you aren't actually getting a property from the class but something it refers to. I personally think it's quite neat, and the 'get' is surely implied. I also think of it slightly more like a builder design, where the constructors for elements are kind of "building" how the behavior of the element is defined. I can meet half way and do ``getInt`` instead of ``getIntAttribute`` which is imo pointlessly verbose. I still like the nouns though.

And the toggle button is just an extension of a button as a convenience class for a state switcher - this is already done manually in a few places throughout structurize and mcol. For example the shape tool has a button to switch between solid and hollow. This is just a cleaner way to define that same button.

Nightenom:
If we are going in the way of modular elements then we can just use brand new modular system (you are particularly right that this is half-way solution). I just like the way it is because: 1) the only problem is buttons 2) classes are not nearly empty 3) you don't have to call getTextModule/getTextureModule to set it's params. Although I see your point and probably even agree with it

Yep, agree on attribute being an unnecessary word in the name, however params.numeral(..) is quite weird against params.getInteger(..)

So it's choice button, I would like to also have checkbox

src/main/java/com/ldtteam/blockout/Color.java
ShadowProtocol: Parsers need their result to be null if the default is to get used when calling getProperty
I was giving people the choice
``@Nullable``

src/main/java/com/ldtteam/blockout/Loader.java
ShadowProtocol: For what exactly? The only thing here is that buttonimage is effectively deprecated

src/main/java/com/ldtteam/blockout/PaneParams.java
ShadowProtocol: I can't actually remember why I did this, but I think there was a semi-sensible reason. I'll look into it and revert if its unimportant.

src/main/java/com/ldtteam/blockout/properties/Tooltip.java
ShadowProtocol: This was the part I did just as you asked for a look, so it is not properly connected. 

Yeah there is a variable mismatch, and tooltip.draw is to be called in drawLast as of this PR, so you can change to use the z stuff at another point

And a tooltip is most definitely an extension imho, as tooltips are never a stand alone concept, always directly represented by the mouse events of a different pane, and they are not part of the standard hierarchy, being at the highest z-index at all times..

src/main/java/com/ldtteam/blockout/views/DropDownList.java
ShadowProtocol: I've seen it used before for fetching first elements, but it is more intuitive when you consider that the same function is used for text with only one line.

src/main/java/com/ldtteam/blockout/Loader.java
Nightenom: there was in the old code

Nightenom:
Ok, let's do it like this:
You want to have functionality modules
I want the classes to have first level methods (like Text.setText(), not Text.text.get().add())

So let's create interfaces like ITooltipModule, ITextModule, ITextureModule (can be a different name, not exactly ...Module) with the old methods name in AbstractText class (so we don't have to change entire mcol code) and getModule() abstract method with return Module as the return type

for text it would look like
```
public RichText getTextModule();
public void setText(String)
{ getTextModule().setText(String); }
```
etc, the implementing class would just have to add the module field and link it to getModule()

edit: for button having three texture modules we still would have to make the methods three times

ShadowProtocol:
Sorry, I'm probably quite the pain by now ... but here we go ...

> I want the classes to have first level methods (like Text.setText()

What's the rationale behind this? If it is backwards compat, we use a specific version of structurize in mcol anyway, I'm happy to adapt changes, the only thing anyone else would have to cope with is a potentially large diff.
Using first-level getters is just not very DRY, and this PR doesn't avtually vary access control that much with ``label.setText()`` vs ``label.text.set()``

That said, I could be convinced to use them like shown in that code block. Maybe.

> So let's create interfaces

Not sure what you're getting at, but I don't think interfaces are suitable for this design. Just use the actual module classes.

> for button having three texture modules we still would have to make the methods three times

I've noticed we are using the enabled-hovered-disabled thing a lot, there might be a better way to represent this.

And yes checkboxes were on my list, as are a couple of other elements. I wanted to get familiar with the library first and any adjustments before adding more (the toggle was essentially a practice).

src/main/java/com/ldtteam/blockout/Loader.java
Raycoms: register("label", Text.class); // TODO: remove, but we don't want to deal with xml changes now

He means this todo

src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -29,92 +30,90 @@
...
+        register("treeview", TreeView::new);
     }
 
-    private Loader()
+    /** The maximum numer of cached documents */
+    private static final int CACHE_CAP = 20;
+
+    /** A map to store the parsed documents. Retains data based on a priority */
+    private static final Map<ResourceLocation, Tuple<Integer,PaneParams>> parsedCache = Collections.synchronizedMap(new HashMap<ResourceLocation, Tuple<Integer, PaneParams>>()
     {
```
Raycoms, line 61: do we need a synchronized map here? Not better to use a caffeeine cache then?

src/main/java/com/ldtteam/blockout/views/DropDownList.java
Raycoms: wouldn't a text.get(0) be more intuitive?

Nightenom:
> the only thing anyone else would have to cope with is a potentially large diff. 

The biggest one, also ppl need to read/look for new mechanisms

Instead of rewriting this from scratch we should focus on the new blockout and expand it by things in this one.
I'm leaving this up to you, the only thing I want is to not change the face of this blockout (ie. no breaking changes unless needed), regarding the modularity I'm ok with it except the previous note

Raycoms:
> > the only thing anyone else would have to cope with is a potentially large diff.
> 
> The biggest one, also ppl need to read/look for new mechanisms
> 
> Instead of rewriting this from scratch we should focus on the new blockout and expand it by things in this one.
> I'm leaving this up to you, the only thing I want is to not change the face of this blockout (ie. no breaking changes unless needed), regarding the modularity I'm ok with it except the previous note

The problem is that the new one is very far from being a) ready and b) mature and c) Incredibly difficult to work with compared to this one.

Nightenom:
That's why we should waste time on it rather than on this one :D

Raycoms:
I mean it's all open source, so everyone can work on what they think is the most worthy to get improvements. The improvements here directly affects what we have in Minecolonies. While the other one may affect Minecolonies much further down the road.

src/main/java/com/ldtteam/blockout/Loader.java
ShadowProtocol: Not completely sure, but because windows are ``open()``ed right after they're constructed I think it does have to be synced.

src/main/java/com/ldtteam/blockout/views/DropDownList.java
ShadowProtocol: The code allows for that, but again this is clarity for multi-line vs single-line. This way implies that I only ever expect to work with the first line. But yes can add the index if you prefer.

ShadowProtocol:
I am more than happy to expand, improve and work with v2 if someone (**a**) makes it buildable - it's currently broken, at least last time I tried it (**b**) consolidates the package sprawl, and (**c**) documents it better, or at least explain to me what's happening. 
I would do all of this myself, but it will be 10x faster if somebody who worked on it does it (or someone familiar with the design methodology).

If that happens sometime soon, then IMHO we shouldn't be upgrading v1 at all to incentivize the transfer (i.e. porting only). If that's not going to happen for a year or more, if ever, then I think no effort should be made to prevent breaking changes in v1.

ShadowProtocol:
Reversion complete. Have at it.

Also I ran IDEA's import optimizer on the package somewhere along the line, so don't mind all the semi-random import adjustments.

src/main/java/com/ldtteam/blockout/Color.java
```diff
@@ -54,21 +70,39 @@ public static int getByName(final String name, final int def)
...
      */
-    public static int rgbaToInt(final int red, final int green, final int blue, final int alpha)
+    public static int rgbaToInt(final int r, final int g, final int b, final int a)
     {
-        int color = alpha;
-        color = (color << 8) + red;
-        color = (color << 8) + green;
-        color = (color << 8) + blue;
-        return color;
+        return (a << 24) | (r << 16) | (g << 8) | b;
```
Nightenom: Just safety check
```suggestion
        return ((a & 0xff) << 24) | ((r & 0xff) << 16) | ((g & 0xff) << 8) | (b & 0xff);
```

src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -29,92 +30,90 @@
...
+        register("dropdown", DropDownList::new);
+        register("overlay", OverlayView::new);
+        register("gradient", Gradient::new);
+        register("zoomdragview", ZoomDragView::new);
+        register("treeview", TreeView::new);
     }
 
-    private Loader()
+    /** The maximum number of cached documents */
+    private static final int CACHE_CAP = 2;
```
Nightenom: Probably client config, not sure about the default value tho (might be worth increasing it).

src/main/java/com/ldtteam/blockout/Parsers.java
```diff
@@ -0,0 +1,180 @@
...
+    public static Function<String, IFormattableTextComponent> TEXT = v -> {
+        String result = v == null ? "" : v;
+        Matcher m = Pattern.compile("\\$[({](\\S+)[})]").matcher(result);
+
+        while (m.find())
+        {
+            String translated = LanguageHandler.translateKey(m.group(1));
+            if (translated.equals(m.group(1)))
+            {
+                translated = NO_TRANSLATION;
```
Nightenom: Nope, just leave the text as it is, would be hard to debug what is the actual missing key

src/main/java/com/ldtteam/blockout/Pane.java
```diff
@@ -1018,4 +1003,18 @@ protected static void blitRepeatable(final MatrixStack ms,
         RenderSystem.enableAlphaTest();
         WorldVertexBufferUploader.draw(buffer);
     }
+
+    public static void fillGradient(
```
Nightenom: Probably not needed

src/main/java/com/ldtteam/blockout/Parsers.java
Nightenom: We had 
image

Nightenom (approved):
Few nitpicks, will merge next weekend (after hopefully last exam) and test something, then merge master

src/main/java/com/ldtteam/blockout/Loader.java
ShadowProtocol: Whoops, that was just for testing. Somewhere between 10 and 30 probably. I'm just trying to think of how many GUIs people access in an hour.

src/main/java/com/ldtteam/blockout/Pane.java
ShadowProtocol: A leftover from the reversion.

src/main/java/com/ldtteam/structurize/config/ClientConfiguration.java
```diff
@@ -24,5 +29,6 @@
     protected ClientConfiguration(final ForgeConfigSpec.Builder builder)
     {
         excludedEntities = defineList(builder, "excludeEntities", new ArrayList<>(), s -> s instanceof String && ResourceLocation.tryCreate((String) s) != null);
+        windowCacheCap = defineInteger(builder, "windowCacheCap", 12, 0, 100);
```
Nightenom, line 32: missing translations

Nightenom (approved):

src/main/resources/assets/structurize/lang/en_us.json
```diff
@@ -1154,5 +1154,7 @@
   "structurize.gui.replaceblock.ambiguous_properties": "Transformation from %s to %s is ambiguous because the following properties are not present in the source block: %s",
   "structurize.config.excludeentities": "Entities Excluded from Blueprint Rendering",
   "structurize.config.excludeentities.comment": "Excludes entities from blueprint rendering if they can't load clientside. If you are going to add a \"minecraft\" entity or you believe it's our fault that an entity couldn't be set up, please make an issue at our GitHub and add the stacktrace to it.",
+  "structurize.config.windowcachecap": "Max Cached GUI Windows",
+  "structurize.config.windowcachecap.comment": "Sets the maximum number of parsed gui window files to be stored for quick loading.",
```
ravenbuilder934:
```suggestion
  "structurize.config.windowcachecap.comment": "Sets the maximum number of parsed GUI window files to be stored for quick loading.",
```

ShadowProtocol: I should've known lazily typing a single translation line, for a comment no less, would invoke the Raven.

ravenbuilder934: i mean obviously :P

ravenbuilder934:
Is this ready to be merged?

Nightenom:
> Is this ready to be merged?

During weekend

PR #278, opened 2021-02-02
Two caches

Raycoms:
- Separates into a dynamic and a static cache
- Have to test this with minecolonies yet
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
+++ b/src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
@@ -8,6 +8,7 @@ import it.unimi.dsi.fastutil.ints.Int2LongMap;
 import it.unimi.dsi.fastutil.ints.Int2ObjectArrayMap;
 import net.minecraft.client.Minecraft;
 import net.minecraft.util.math.BlockPos;
+
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
```diff
@@ -19,12 +20,15 @@
     /**
      * A static instance on the client.
      */
-    private static final BlueprintHandler ourInstance = new BlueprintHandler();
-    private static final int CACHE_SIZE = 30;
-    private static final long CACHE_EVICT_TIME = 45_000L;
+    private static final BlueprintHandler ourInstance      = new BlueprintHandler();
+    private static final int              CACHE_SIZE       = 30;
```
Nightenom, line 24: Use smaller size for static and something like 15sec (15000ms) evict time. So it doesn't eat so much

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
```diff
@@ -83,16 +109,29 @@ public void draw(final Blueprint blueprint, final BlockPos pos, final MatrixStac
      */
     public void cleanCache()
     {
-        final long now = System.currentTimeMillis();
-        final Iterator<Int2LongMap.Entry> iter = evictTimeCache.int2LongEntrySet().iterator();
+        long now = System.currentTimeMillis();
```
Nightenom, line 112: we can only use one "now"

Nightenom (approved):
Remove changes from other PRs

src/main/java/com/ldtteam/structures/client/BlueprintHandler.java
Raycoms: I might be "overdoing it", wondered if so much time passed after all of the above, to create a new one lower.

PR #280, opened 2021-02-03
Add render source

Raycoms:
Closes #
Closes #
Closes #

I talked a bunch with Orion and Someaddons about this. I think this is the only true solution to this problem.
On the Minecolonies side, for static schematics (that do not change their position), we'll set the source position in the blueprint to be included in the hash. This will solve the problem completely and will avoid any interference

- Include Render Source in Blueprint hash
-
-

Review please


Nightenom (approved):
>I think this is the only true solution to this problem.
 
Definitely not a true solution, but sure go ahead

Raycoms:
The one true solution that doesn't take a rewrite of structurize xD

PR #283, opened 2021-02-04
Adjustments

Nightenom:
- fixes to latest blockout upgrade

Review please


src/main/java/com/ldtteam/blockout/controls/AbstractTextElement.java
```diff
@@ -396,20 +403,39 @@ public void setText(final List<IFormattableTextComponent> text)
...
+    }
+
     public void setText(final IFormattableTextComponent text)
     {
-        this.text = Collections.singletonList(text);
-        recalcTextRendering();
+        setText(Collections.singletonList(text));
+    }
+
+    public void clearText()
```
Raycoms, line 445: just missing javadoc on those

Raycoms:
:shipit:

Nightenom:
I might add more as I go through mcol code :D

src/main/java/com/ldtteam/blockout/Loader.java
```diff
@@ -275,4 +280,9 @@ public static void addToCache(ResourceLocation loc, PaneParams doc)
 
         parsedCache.put(loc, new Tuple<>(1, doc));
     }
+
+    public static void cleanXMLparseCache()
```
Raycoms: jdoc

src/main/java/com/ldtteam/blockout/controls/AbstractTextElement.java
```diff
@@ -268,6 +294,8 @@ else if (textAlignment.isHorizontalCentered())
             lineShift += mc.fontRenderer.FONT_HEIGHT + textLinespace;
         }
 
+        // TODO: forge disable filtering
```
Raycoms, line 300: todo for the future?

Nightenom: todo for this <https://github.com/MinecraftForge/MinecraftForge/pull/7645>

src/main/java/com/ldtteam/blockout/controls/ToggleButton.java
```diff
@@ -44,8 +76,32 @@ public ToggleButton(final PaneParams params)
...
+    /**
+     * @param raw true if the result should not be localized
+     * @return the list of states as strings
+     */
+    public List<String> getStateStrings(boolean raw)
+    {
+        return raw ? rawStates : states.stream().map(IFormattableTextComponent::getString).collect(Collectors.toList());
     }
 
+    public List<IFormattableTextComponent> getStates()
```
Raycoms, line 94: jdoc

src/main/java/com/ldtteam/blockout/controls/AbstractTextElement.java
Raycoms: kk

Raycoms:
Looks fine for me, only missing some javadoc here and there :shipit: 

@OrionDevelopment @ShadowProtocol

src/main/java/com/ldtteam/blockout/controls/ToggleButton.java
ShadowProtocol: We haven't been adding jdoc for pure getters like this one. It wouldn't really clarify anything further. Between all the getters around that's hundreds of lines of code we don't really need. There was a brief discussion on this in the original PR. This is what I said then:
> I just imagine it as if it is in an auto-genned doc page. would someone understand that this sets the text alignment based off of the method title alone? If the answer is yes then it doesn't need a jdoc IMO. You just need to be careful that you aren't doing other things in the setter that won't be described by the title alone. 
But if you're just going to chuck ``/** Sets the text alignment */`` above it then that is not worth anyone's energy scrolling past it.

But if you insist then I will.

PR #289, opened 2021-02-09
Block Grouping, Registration, and Generation

ShadowProtocol:
- All blocks and items are now registered using DeferredRegistries, more suitable for making add-ons.
- Introduced Block Collections to rapidly create custom sets of blocks, and make data gen easier
    - Is generic, so can be used by dependents 
    - New types can be added in a single enum and gen call
- Significantly reduced the number of block definition files
- Added a set of data providers to generate data for collections
- Introduce a mod-wide language provider that will sort the lang keys to reduce future diff size
- Adds a creative group for schematics construction blocks
- Proposal to adjust some of the timber frame translations slightly (this may be a tad confusing for veteran players though)

Just a tad rough round the edges, needs a smidge of docing still, and haven't quite worked out how the tag providers work without overriding the previous providers efforts

Something to work on (whether in this PR or not) is utilizing ``IBlockList`` in other blocks like shingles, to register and generate them even more efficiently (already pretty good though)

Review please


src/main/java/com/ldtteam/structurize/blocks/ModBlocks.java
```diff
@@ -1,274 +1,171 @@
...
-    public static BlockFluidSubstitution blockFluidSubstitution;
+    public static DeferredRegister<Block> getRegistry()
+    {
+        return BLOCKS;
+    }
 
-    /**
-     * Brown, beige, and cream bricks (and variants)
+    /*
+     *  Lone Blocks
```
ravenbuilder934, line 47:
```suggestion
     * Lone Blocks
```
Extra space here

src/main/java/com/ldtteam/structurize/blocks/ModBlocks.java
```diff
@@ -1,274 +1,171 @@
...
+        RegistryObject<B> registered = BLOCKS.register(name.toLowerCase(), block);
+        ModItems.getRegistry().register(name.toLowerCase(), () -> new BlockItem(registered.get(), new Item.Properties().group(group)));
+        return registered;
+    }
 
-        blockBrownBrickStairs = new BlockBrickStairs("blockbrownbrickstairs").registerBlock(registry);
-        blockBeigeBrickStairs = new BlockBrickStairs("blockbeigebrickstairs").registerBlock(registry);
-        blockCreamBrickStairs = new BlockBrickStairs("blockcreambrickstairs").registerBlock(registry);
+    /*
+     *  Registration
```
ravenbuilder934, line 126: Extra space(s) here too

src/main/java/com/ldtteam/structurize/blocks/types/TimberFrameType.java
```diff
@@ -7,15 +7,15 @@
...
+    PLAIN("plain", "Vertical", false),
+    DOUBLE_CROSSED("double_crossed", "Crossed", false),
+    FRAMED("framed", "Plain", false),
     SIDE_FRAMED("side_framed", "Side Framed", true),
-    UP_GATED("up_gated", "Up Gate Framed", true),
-    DOWN_GATED("down_gated", "Down Gate Framed", true),
-    ONE_CROSSED_LR("one_crossed_lr", "Left Right Crossed", false),
-    ONE_CROSSED_RL("one_crossed_rl", "Right Left Crossed", false),
-    HORIZONTAL_PLAIN("horizontal_plain", "Plain Horizontal", false),
+    UP_GATED("up_gated", "Up-gate", true),
```
ravenbuilder934:
```suggestion
    UP_GATED("up_gated", "Up-Gate", true),
```
For down-gate, right-hand, and left-hand too

ravenbuilder934 (commented):
Couple nitpicks
Also, how is this for compat with my nether blocks PR? Would that one need to start again from scratch if this was merged?

src/main/java/com/ldtteam/structurize/blocks/types/TimberFrameType.java
ShadowProtocol: They're the U-shaped ones, they don't have a left-hand right-hand

src/main/java/com/ldtteam/structurize/blocks/ModBlocks.java
ShadowProtocol:
```java
/*
 * This is a comment. And if we are all honest with each other, 
 * comments are likely written quickly, and probably as an afterthought.
 * Whitespace, punctuation and formatting differ greatly across the board, and will 
 * likely continue to do so for aeons to come.
 */
 ```
 It's a tab-space, and I'll have you note it lines right up with the code above it.

**Comment by ShadowProtocol** [comment]
Probably easiest for you to rebase onto this for those edits, or merge after this has.

File: src/main/java/com/ldtteam/structurize/network/NetworkChannel.java [hunk 4]
```diff
@@ -60,7 +61,6 @@ public void registerCommonMessages()
         registerMessage(++idx, StructurizeStylesMessage.class, StructurizeStylesMessage::new);
         registerMessage(++idx, UndoMessage.class, UndoMessage::new);
         registerMessage(++idx, UpdateScanToolMessage.class, UpdateScanToolMessage::new);
-        registerMessage(++idx, UpdatePlaceholderBlockMessage.class, UpdatePlaceholderBlockMessage::new);
```
Nightenom, line 63: Even though we usually don't allow running different structurize version at client and server, please keep the idx increment and add todo remove 1.17 comment

Nightenom (commented):
I will go through later cuz exams, but have you took a look into if it's possible to sort JsonObject while serializing to json? because right now it's random and causes git diff to explode each time we run datagen (I think it's possible but haven't got to it yet)

edit: regarding the lang file it would be the best if we can move the datagen things to the end of file, probably can use something like `"__comment":"data_gen"` to make a border, sorting it alphabetically doesn't make it any better than it's now

src/main/java/com/ldtteam/structurize/network/NetworkChannel.java
Raycoms: Nah, we do this regularly in minecolonies. I don't have a problem with this.

src/main/java/com/ldtteam/structurize/blocks/types/TimberFrameType.java
ravenbuilder934:
```java
    DOWN_GATED("down_gated", "Down-gate", true),
    ONE_CROSSED_LR("one_crossed_lr", "Right-hand Slope", false),
    ONE_CROSSED_RL("one_crossed_rl", "Left-hand Slope", false),
```
I meant these

src/main/java/com/ldtteam/structurize/blocks/IBlockCollection.java
```diff
@@ -0,0 +1,182 @@
...
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.fml.RegistryObject;
+import net.minecraftforge.registries.DeferredRegister;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public interface IBlockCollection
```
Raycoms: This def needs javadoc

src/main/java/com/ldtteam/structurize/generation/LanguageWriter.java
```diff
@@ -0,0 +1,134 @@
...
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.TreeMap;
+import java.util.function.Function;
+
+public final class LanguageWriter implements IDataProvider
```
Raycoms: Javadoc

src/main/java/com/ldtteam/structurize/generation/DataGeneratorConstants.java
```diff
@@ -45,9 +46,13 @@
...
-        if (jsonElement.isJsonObject())
-            jsonElement.getAsJsonObject().addProperty("READ ME!!!", "DO NOT EDIT THIS FILE. DON'T TOUCH IT, TOUCH IT AND YOU WILL REGRET IT ~~ AshersLab (Use the Data Generators!");
-        return jsonElement;
+        return serialize(serializable.serialize());
+    }
+
+    public static JsonElement serialize(JsonElement serialized)
+    {
+        if (serialized.isJsonObject())
+            serialized.getAsJsonObject().addProperty("READ ME!!!", "DO NOT EDIT THIS FILE. DON'T TOUCH IT, TOUCH IT AND YOU WILL REGRET IT ~~ AshersLab (Use the Data Generators!");
```
Asherslab: the missing closing bracket (that i failed to put in the first place) is bugging me here =P

src/main/java/com/ldtteam/structurize/blocks/decorative/BlockTimberFrame.java
```diff
@@ -100,7 +98,7 @@ public TimberFrameFrameType getFrameType()
      *
      * @return the registered TimberFrameCentreType
      */
-    public TimberFrameCentreType getCentreType()
+    public TimberFrameCenterType getCenterType()
```
Asherslab: it's spelt Centre outside the US: https://dictionary.cambridge.org/dictionary/english/centre

Asherslab (changes requested):
Looks good otherwise. Glad we could removed our Abstract base classes.

src/main/java/com/ldtteam/structurize/blocks/decorative/BlockTimberFrame.java
ShadowProtocol: I know! I *am* outside the U.S! But unfortunately they bet us to mass collaborative programming, and now they inflict their spelling conventions on us all. It was a toss-up, but I decided I would rather look at conventional spelling than proper spelling.

Asherslab: it was fine as is, just don't change it lol

src/main/java/com/ldtteam/structurize/generation/DataGeneratorConstants.java
ShadowProtocol: I think you mean "(that I failed to put on in the first place"

Asherslab: damn you

ShadowProtocol:
> I will go through later cuz exams, but have you took a look into if it's possible to sort JsonObject while serializing to json? because right now it's random and causes git diff to explode each time we run datagen (I think it's possible but haven't got to it yet)

Oh that's simple (I think), we're using ``HashMap``s everywhere in the datagenerators dependency. We either want a ``TreeMap`` or ``LinkedTreeMap``, or add each property from a list of a class that stores the key and values somehow. I've noticed Forge's builders use HashMaps too. Worth bringing to their attention?
> edit: regarding the lang file it would be the best if we can move the datagen things to the end of file, probably can use something like `"__comment":"data_gen"` to make a border, sorting it alphabetically doesn't make it any better than it's now

This was actually a solution to the above problem, we change that and this will be easily done.

ravenbuilder934 (commented):
You changed the names of the brick textures, but as far as I can tell you didn't change the references to their old names

ShadowProtocol:
> You changed the names of the brick textures, but as far as I can tell you didn't change the references to their old names

Sure did! It was all moved into code for the generators, kinda what the whole PR is for. All the current jsons are sitting there useless. As preferred, I haven't run the generator in this PR, but once I have you'll see some new jsons with the right references.
Good reminder to just delete the current ones though.

ravenbuilder934:
ah, the generators not getting run yet was confusing me
👍

Asherslab:
> I haven't run the generator in this PR, but once I have you'll see some new jsons with the right references.

just incase you didn't see the messages about it. i believe the preferred way to do that is create a PR into this PR

ShadowProtocol:
Yep, I'm still figuring out the tag side of things, so I'm holding off.

Since you're here, is it worth a PR to forge to replace all their hash maps with deterministic ones? Kinda makes the diff large for every mod that uses the provider builders.

Raycoms:
> Yep, I'm still figuring out the tag side of things, so I'm holding off.
> 
> Since you're here, is it worth a PR to forge to replace all their hash maps with deterministic ones? Kinda makes the diff large for every mod that uses the provider builders.

Checked with some people and it seems it's a good idea

Asherslab:
> Checked with some people and it seems it's a good idea

confirmed with Forge, lex, and Shadow, they're already deterministic and that Shadow is just going nuts

ShadowProtocol:
> Shadow is just going nuts

This is true. 
Alright, with some tweaking I think I can remove our dependency on the datagenerators project entirely.

Asherslab:
>  I think I can remove our dependency on the datagenerators project entirely.

nice, was a temporary-became-permanent solution anyway

Raycoms:
@Nightenom @Asherslab @ShadowProtocol what's missing here?

ShadowProtocol:
I'm reworking a couple of things, mainly to make the generation more API-friendly. I came across an issue with overriding some files across providers, so I've spent some time to make only one single provider of each type be called, but it changes the design pattern a bit. Good news is it should be easy for mcol to hook on to, and culls even more repetitive code.

I don't think I'm to far away from being ready to more formally review.

Raycoms:
> I'm reworking a couple of things, mainly to make the generation more API-friendly. I came across an issue with overriding some files across providers, so I've spent some time to make only one single provider of each type be called, but it changes the design pattern a bit. Good news is it should be easy for mcol to hook on to, and culls even more repetitive code.
> 
> I don't think I'm to far away from being ready to more formally review.

Sounds good

ShadowProtocol:
Alright, it's in a working state, though there are a couple of oddities that arise from using the built-in providers. I've found them to be really helpful, but somewhat restrictive. One example is we're forced to create criterion for every single recipe, that adds them all to the vanilla recipe book, and toasts every time you find a new one. And there are other more subtle disadvantages too, but I think they're offset by the additions to deletions ratio.

src/api/java/com/ldtteam/structurize/api/blocks/BlockType.java
```diff
@@ -0,0 +1,149 @@
...
+    FENCE_GATE("fence_gate", FenceGateBlock::new, BlockTags.FENCE_GATES, ItemTags.FENCES, 1, "-#-", "-#-"),
+    TRAPDOOR("trapdoor", TrapDoorBlock::new, BlockTags.TRAPDOORS, ItemTags.TRAPDOORS, 3,  "###", "###"),
+    DOOR("door", DoorBlock::new, BlockTags.DOORS, ItemTags.DOORS, 3, "##", "##", "##");
+
+    public final  String                               suffix;
+    public final Function<Properties, ? extends Block> constructor;
+    public final List<ITag.INamedTag<Block>>           blockTag;
+    public final List<ITag.INamedTag<Item>>            itemTag;
+    private final int                                  recipeYield;
+    private final String[]                             patterns;
```
Nightenom: maybe rename to crafting pattents, would also rather see fixed amount of entries or list

src/main/generated/resources/assets/structurize/lang/en_us.json
```diff
@@ -0,0 +1,3 @@
+{
```
Nightenom: Does it really need to exist? should be created on the fly if entire generated folder get deleted

src/main/java/com/ldtteam/structurize/Structurize.java
```diff
@@ -40,6 +43,8 @@
      */
     public Structurize()
     {
+        ModBlocks.getRegistry().register(FMLJavaModLoadingContext.get().getModEventBus());
+        ModItems.getRegistry().register(FMLJavaModLoadingContext.get().getModEventBus());
```
Nightenom: nitpick: move it below the config line to other event registers

src/main/java/com/ldtteam/structurize/client/gui/WindowAddTag.java
```diff
@@ -1,47 +0,0 @@
-package com.ldtteam.structurize.client.gui;
```
Nightenom, line 1: Hmm?

src/main/java/com/ldtteam/structurize/client/gui/WindowPlaceholderblock.java
```diff
@@ -1,283 +0,0 @@
-package com.ldtteam.structurize.client.gui;
```
Nightenom, line 1: Also this? reason for removing unrelated classes?

src/main/java/com/ldtteam/structurize/generation/StructurizeTranslations.java
```diff
@@ -0,0 +1,133 @@
...
+        add(ModBlocks.blockFluidSubstitution.get(), "Fluid Placeholder");
+        add(ModItems.caliper);
+        add(ModItems.buildTool.get(), "Build Tool");
+        add(ModItems.scanTool.get(), "Scan Tool");
+        add(ModItems.tagTool.get(), "Tag Tool");
+        add(ModItems.shapeTool.get(), "Shape Tool");
+        add(ModItemGroups.STRUCTURIZE, "Structurize");
+        add(ModItemGroups.CONSTRUCTION, "Structurize Construction Blocks");
+        add(ModItemGroups.SHINGLES, "Structurize Shingles");
+        add(ModItemGroups.TIMBER_FRAMES, "Structurize Timber Frames");
```
Nightenom: veto from me, (feel free to override, but give reasons)
non-generated things like gui and tools should be in json and not code

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
```diff
@@ -233,7 +230,7 @@ private void load(final String path)
             InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(String.format(path, locale));
             if (is == null)
             {
-                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(String.format(path, defaultLocale));
+                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(String.format("../generated/resources/" + path, defaultLocale));
```
Nightenom: This will crash on mcol, move to ctor

Nightenom (changes requested):
Before merge, I want to know that it produces the same jsons as we have now :)
I like having block registry name in its class, but don't care
Some formatting nitpicks like using final and {} on the same line with comment (private ctors)
I saw you removed the placeholder block so don't mind those comments around it
(edit: btw not sure if it's good to remove it now, maybe wait for 1.17)

Discussion needed around lang file:
a) should be crowdin friendly
b) generated things can be in the same file but separated
c) I veto non-generated things being in code because of a) and maintainability

To describe a) more:
There is a problem for translating the massive generation mess, the solution is quite simple: have standalone words as keys, have format keys for block names (probably needs MessageFormat (or what's class name) which support ordering of keys, eg. {1} {0})
Then we either generate the translation on the fly during mod setup (we can cache the result in our structurize folder) or generate them during CI build after crowding task

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
Nightenom: Also this is mostly runtime class, so probably needs to distinguish dev space and jar space

src/api/java/com/ldtteam/structurize/api/generation/ModBlockStateProvider.java
```diff
@@ -0,0 +1,117 @@
...
+            {
+                return new ResourceLocation(name.getNamespace(), directory + "/" + name.getPath());
+            }
+        }
+
+        // Take the last one when none are found. It gives a clearer error message
+        return new ResourceLocation(name.getNamespace(), directory + "/" + name.getPath());
+    }
+
+    public void stairsBlockUnlockUV(StairsBlock block, ModelFile stairs, ModelFile stairsInner, ModelFile stairsOuter) {
```
Raycoms: formatting off in this method

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
ShadowProtocol: Need to track locale though, so I used a flag.

Not quite sure what you want to do for the latter.

src/main/java/com/ldtteam/structurize/generation/StructurizeTranslations.java
Raycoms: Agreed

src/main/generated/resources/assets/structurize/lang/en_us.json
ShadowProtocol: Oops this was an artifact from a previous experiment.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
Nightenom: There is dev space flag FMLEnvironment#production (true for jar space, false for dev space)

src/main/java/com/ldtteam/structurize/generation/StructurizeTranslations.java
ShadowProtocol: This was me following direction from Asher in a brief discussion we had.
It was a toss-up between this option, and having the provider read and copy the manual one to the generated one.
Between these two options, there should be no reason to keep the manual and genned keys in the same file. 

For the record I preferred the latter one, but this one is growing on me, for the following advantages:
 - Using these add overloads it is impossible to accidentally misspell a block or item key
 - You can use autocomplete in your IDE for the variable based overloads
 - The order in any lang file doesn't really matter, as it can be determined here.
 - Allows easy access to the provider should anything dynamic be needed that isn't covered by the collection/list interfaces
 - Once more overloads are added (for 1.17) it will begin to enforce consistency. 
    - Done correctly this will *increase* maintainability, or effectively remain the same.
 
I haven't looked deeply into CrowdIn, but if you can't just switch the input to the generated file I'd be very surprised. No need to mess with code at all.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
ShadowProtocol: Alright. But what needs to be different? Both need to read the genned lang file for Structurize.

Nightenom: In jar space you don't have generated folder at all, the assets folder sits in the root of jar

ShadowProtocol: oh duh, I forgot about that.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
```diff
@@ -235,7 +236,8 @@ private void load(final String path, final boolean generatedDefault)
             InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(String.format(path, locale));
             if (is == null)
             {
-                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(String.format((generatedDefault ? "../generated/resources/" : "") + path, defaultLocale));
+                is = Thread.currentThread().getContextClassLoader().getResourceAsStream(
+                  String.format((generatedDefault && !FMLEnvironment.production ? "../generated/resources/" : "") + path, defaultLocale));
```
Nightenom: You could just do at line 219 load(fileLoc, !FMLEnvironment.production)

Nightenom: Try to make it so it works automatically without the generatedDefault boolean flag

src/main/java/com/ldtteam/structurize/generation/StructurizeTranslations.java
Nightenom: - regarding IDE and overloads, you can still fail doesn't matter (guis/commands etc.), we will just add more overloads over time making it even less maintainable, also - the biggest drawback - reloads, you can't reload or it's super annoying to reload genned files during debug

- the order does matter for both, genned needs to be sorted cuz of git, manual is usually "foldered" by usage type etc.

- crowdin is irrelevant in this problem (but the problems above still applies), you can merge jsons in gradle (for build), you can load multiple jsons in dev space (even into mc language handler)

ShadowProtocol: - The overloads are one-liners, I would hardly consider that to be a maintainability issue until you get to 20 or so of them. You make a good point about reload though - reloading/regenning time was why this PR took so long!
- I guess what I was trying to get at was the "foldered" style is a bit more visually apparent due to syntax highlighting. But its a subtle advantage, and depends how you look at it I guess.
- I did not know gradle could do that. And the multiple jsons was what I asked in Discord that started this! So if you know how to do it I'm quite happy to use that method. I'll just blame Asher.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
ShadowProtocol: I was just trying to account for the possibility that minecolonies might want to also use genned lang sometime.

Nightenom: But it still would be based on in which space you are

src/main/java/com/ldtteam/structurize/generation/StructurizeTranslations.java
Nightenom: > I did not know gradle could do that. And the multiple jsons was what I asked in Discord that started this! So if you know how to do it I'm quite happy to use that method. I'll just blame Asher.

google it :D gradle is "basically" another semi-java code so you can just import classes for json and do whatever you want

build.gradle
```diff
@@ -48,6 +50,7 @@ sourceSets {
         resources {
             srcDir 'src/main/resources'
             srcDir 'src/main/generated/resources'
+            exclude "assets/structurize/lang/default.json"
```
Nightenom, line 53: shouldn't we exclude the manual file too?

build.gradle
```diff
@@ -376,6 +379,22 @@ task setupCIWorkspace {
     }
 }
 
+processResources {
+    onlyIf {
+        // this won't exist on the first time running data gen after wiping the generated folder
+        file('src/main/generated/resources/assets/structurize/lang/default.json').exists()
```
Nightenom, line 385: hoping the onlyIf is only for the doLast part and not entire processResources

src/main/java/com/ldtteam/structurize/blocks/decorative/BlockTimberFrame.java
```diff
@@ -1,26 +1,21 @@
...
 import net.minecraft.state.StateContainer;
-import net.minecraft.state.properties.BlockStateProperties;
 import net.minecraft.util.Mirror;
-import net.minecraft.util.Rotation;
 
 /**
  * Decorative block
  */
-public class BlockTimberFrame extends AbstractBlockStructurize<BlockTimberFrame>
+public class BlockTimberFrame extends DirectionalBlock
```
Nightenom, line 17: Is this safe in terms of block data?

src/main/java/com/ldtteam/structurize/generation/DataGeneratorConstants.java
```diff
@@ -45,9 +25,13 @@
...
     public static JsonElement serialize(IJsonSerializable serializable)
     {
-        JsonElement jsonElement = serializable.serialize();
-        if (jsonElement.isJsonObject())
-            jsonElement.getAsJsonObject().addProperty("READ ME!!!", "DO NOT EDIT THIS FILE. DON'T TOUCH IT, TOUCH IT AND YOU WILL REGRET IT ~~ AshersLab (Use the Data Generators!");
-        return jsonElement;
+        return serialize(serializable.serialize());
+    }
+
+    public static JsonElement serialize(JsonElement serialized)
```
Nightenom: Was questioning myself if we should remove this :D would save few bytes

src/main/java/com/ldtteam/structurize/generation/StructurizeTranslations.java
Nightenom: remove this file?

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
```diff
@@ -111,11 +108,6 @@ else if (object instanceof String)
             }
         }
 
-        if (translation == null)
```
Nightenom, line 114: hmm?

Nightenom (commented):
Just few comments
Last thing - is the datagen at least "git stable"?

build.gradle
ShadowProtocol: I save it to the build version of the manual one, so excluding it would leave us with no lang file at all.

build.gradle
ShadowProtocol: Good question - I truly have no idea, but I ran the client and it worked, so I'm assuming the condition only applies to this closure

src/main/java/com/ldtteam/structurize/blocks/decorative/BlockTimberFrame.java
ShadowProtocol: The only difference is that the directional property now exists in the vanilla superclass - so that should not affect block data at all.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
ShadowProtocol: That was an IDEA suggestion and I was feeling carefree. Plus I like to see as little yellow squiggles as possible :)

Nightenom: looks like this line <https://github.com/ldtteam/Structurize/pull/289/files/bbddcaeab9693ba5fc0597d030399f3dc32a4416#diff-6676addc4748aa15f73e5265e569a68c4c70246518b07987e045dafdd427edf1L67> should be > 0 instead of >=0

ShadowProtocol: Sorry that link doesn't redirect to a line, where are you talking about?
*nvm I see it*

Nightenom: https://github.com/ShadowProtocol/Structurize/blob/feature/datagen/src/main/java/com/ldtteam/structurize/util/LanguageHandler.java#L64

ShadowProtocol: Hard to tell what this segment is actually trying to achieve purely from the code, but after a bit of analysis I think it's fine as is.

build.gradle
```diff
@@ -376,6 +379,22 @@ task setupCIWorkspace {
...
+    }
+    doLast {
+        def slurper = new JsonSlurper()
+        def standard = slurper.parse(file('src/main/resources/assets/structurize/lang/en_us.json'))
+        def generated = slurper.parse(file('src/main/generated/resources/assets/structurize/lang/default.json'))
+
+        standard += generated
+
+        file('build/resources/main/assets/structurize/lang/en_us.json').write(JsonOutput.prettyPrint(JsonOutput.toJson(standard)))
```
Nightenom: we can go without pretty print here

ShadowProtocol:
Think this is ready now, Although the difference is so fundamental I wouldn't be surprised if I'd missed something somewhere.

Also when you say 'git stable' what do you mean? This will produce a ***giant*** diff the first time, but all future runs should only reflect actual changes.

Nightenom (approved):
By git stable I mean that running datagen two times with the same data produces exactly the same output - data that git considers to be equal
General approve without looking at the datagen stuff - leaving that to Asher as he probably knows better
Last nitpick I just saw - there are some 1.17 todo comments - I would appreciate **not** changing any block ids unless forced

marchermans (approved):

PR #296, opened 2021-02-20
Fix cursor logic during drawing

Nightenom:
- fix a problematic calling of isPointInPane in different scaling contexts
- replace ItemIcon tooltip renderer with our tooltip

Review please


Raycoms:
:shipit: go with it

PR #300, opened 2021-02-24
Better method to prevent blocks (i.e. fences) from connecting.

Sandriell:

Removes the smaller physics shape/hitbox and overrides getBlockFaceShape to prevent blocks (like fences) from connecting.

Review please


Raycoms:
The PR is not compiling

Sandriell:
Had not realized that this method had been removed in 1.13+ :(

PR #318, opened 2021-03-13
[DATA] Block Grouping, Registration, and Generation

ShadowProtocol:
data gen for #289 

- Fixes the jar-breaking if condition in gradle (note to self: always test the jar!)
- Runs the data generator

Review please (only build.gradle was edited manually)


Nightenom (approved):
Told ya that I don't like the onlyIf :D

PR #321, opened 2021-03-14
Fix tag duplication and stability

ShadowProtocol:
- Fixes the leak that generated extra tags for each frame type unnecessarily
- Data gen is now 100% git stable, for me at least :zany_face: 

Only manual changes are in TimberFrameType and ShingleFaceType

Review please


Nightenom:
Btw I just saw that you don't use formatting here, we use same formatting as in mcol everywhere

PR #336, opened 2021-03-31
Don't overwrite existing schematic names

### LINKED ISSUE #335: Scan filename overwrites DecorationController schematicName
When taking a scan of a structure containing a DecorationController (or really any block implementing IBlueprintDataProvider), the filename of the scan blindly overwrites the schematicName.  This presents a situation where it's nearly impossible to build upgradable schematics without manually editing the DecorationController nbt in the blueprint file after the fact.

Tracing the code, it's because of this line here:
https://github.com/ldtteam/Structurize/blob/a4484f5ca9dbf2118b904f26fc2c863157258f04/src/main/java/com/ldtteam/structurize/items/ItemScanTool.java#L185

Any solution for this problem should handle both the situation of a DecorationController with an empty schematicName as well as the situation of a DecorationController with an existing schematicName.

For context, the specific problem I have is as follows:  I'm trying to scan an upgradable schematic named `techyshishy/blackstone/paths/intersection5`.  The file needs to live at `schematics/ts/blackstone/paths/intersectiont5.blueprint`, the DecorationController needs to have a schematicName of `schematics/techyshishy/blackstone/paths/intersection`, and when I actually make the scan, the scan tool has a character limit, so I need to scan it to just `intersection5`, and then move it into place manually.

TechyShishy:
Currently, the process of running a scan over an area that contains a
DecorationController from minecolonies will overwrite the schematicName
field in the DecorationController with the filename provided in the scan
tool.  This change modifies the process so that DecorationControllers
with existing schematicNames preserve their values, but empty
DecorationControllers will inherit the filename.

Closes #335

- Change the scan tool to preserve existing schematic names in decoration controllers

This is one possible method for fixing this problem, and is sort of the more braindead version of doing so.  It doesn't handle the corner case of a user who doesn't manually edit their DecorationController, and scans the same schematic multiple times with different filenames.  This would result in the DecorationController maintaining the very first filename as a schematicName.  In my opinion, this is a reasonable compromise for a first pass at solving this problem.  Feel free to disagree.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
+++ b/src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
@@ -181,7 +181,8 @@ public class ItemScanTool extends AbstractItemWithPosSelector
             final TileEntity te = world.getTileEntity(anchorPos.get());
             if (te instanceof IBlueprintDataProvider)
             {
-                if (((IBlueprintDataProvider) te).getSchematicName().length() == 0) {
+                if (((IBlueprintDataProvider) te).getSchematicName().length() == 0)
+                {
                     ((IBlueprintDataProvider) te).setSchematicName(fileName);
                 }
                 ((IBlueprintDataProvider) te).setCorners(from.subtract(anchorPos.get()), to.subtract(anchorPos.get()));
--- a/src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
+++ b/src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
@@ -181,7 +181,9 @@ public class ItemScanTool extends AbstractItemWithPosSelector
             final TileEntity te = world.getTileEntity(anchorPos.get());
             if (te instanceof IBlueprintDataProvider)
             {
-                ((IBlueprintDataProvider) te).setSchematicName(fileName);
+                if (((IBlueprintDataProvider) te).getSchematicName().length() == 0) {
+                    ((IBlueprintDataProvider) te).setSchematicName(fileName);
+                }
                 ((IBlueprintDataProvider) te).setCorners(from.subtract(anchorPos.get()), to.subtract(anchorPos.get()));
             }
         }


Raycoms:
Okay let me try to guess the problem here. The problem is that the schematic name in the decoration controller should have the correct path "schematics/decorations/etc" and the schematic name its saved in would be just "bathhouse1" and we also don't want the "1" at the end of this in this case.

I think turning this off for all is bad, I do like having this name data in the other tileEntities (our buildings in minecolonies).
Might need some extra input. But if at all, the decoration controller should probably handle this exception itself and not the scan tool.

someaddons:
the blueprint data provider is supposed to have the name, it doesnt make sense to only save the name where there is no schematic name provided

Raycoms:
Yeah, it might make sense to use a different name field in the decoration controller for the actual name that is displayed and a different one for "its own name".

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
```diff
@@ -181,7 +181,9 @@ public static void saveStructure(
             final TileEntity te = world.getTileEntity(anchorPos.get());
             if (te instanceof IBlueprintDataProvider)
             {
-                ((IBlueprintDataProvider) te).setSchematicName(fileName);
+                if (((IBlueprintDataProvider) te).getSchematicName().length() == 0) {
```
ravenbuilder934: bracket on new line

TechyShishy:
Sorry, a more detailed description of the problem is in the linked issue, but yes, you're correct.  The basic idea is that those two paths are often subsets of each other for good reasons, and the current code makes it very difficult to do so.

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
TechyShishy: Fixed in 83c828b

PR #338, opened 2021-04-02
Implement block remapping due to remapped IDs in an earlier commit.

marchermans:
Implements a block and item remapping system that remaps old worlds to the new ids, using the Forge MissingMapping system.

Several regexes catch the old missing names and then attempt a lookup into the registry with new name to see if it exist and then remap it if it finds one.

See the code for the regexes used.


src/main/java/com/ldtteam/structurize/event/EventSubscriber.java
```diff
@@ -83,4 +97,103 @@ public static void onWorldTick(@NotNull final TickEvent.WorldTickEvent event)
...
+    @SubscribeEvent
+    public static void onMissingBlockMappings(final RegistryEvent.MissingMappings<Block> missingBlockEvent)
+    {
+        final ImmutableList<RegistryEvent.MissingMappings.Mapping<Block>> missingBlocks = missingBlockEvent.getMappings(Constants.MOD_ID);
+        if (missingBlocks.isEmpty())
+        {
+            return;
+        }
+
+        handleMissingMappingsDueToRestructureIn9af7543d7dcefa0bf7b52c2e809d6c8d48b803a6(missingBlocks);
```
Nightenom, line 111: this name should be rather in jdoc (or at least not contain the entire commit hash)

Nightenom (commented):
Overall good, one question - does this persists/works for blueprints (bp)? bp uses vanilla dfu for standalone blocks [link](https://github.com/ldtteam/Structurize/blob/version/1.16.3/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java#L241-L273) and then NbtUtil blockstate read method - does this method uses the forge missing mapping system?

_edit: and we (you) should do something with formatting, because the replacementPatterns looks really bad to me_

src/main/java/com/ldtteam/structurize/event/EventSubscriber.java
marchermans: I don't want them to be in the jdoc.
I made the decision that i wanted in the name because there might be several remaps that we need in the future.

marchermans:
> Overall good, one question - does this persists/works for blueprints (bp)? bp uses vanilla dfu for standalone blocks [link](https://github.com/ldtteam/Structurize/blob/version/1.16.3/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java#L241-L273) and then NbtUtil blockstate read method - does this method uses the forge missing mapping system?
> 
> _edit: and we (you) should do something with formatting, because the replacementPatterns looks really bad to me_

Yes this works, since this remaps the objects in the registry, basically on the lowest level.

What is wrong with formatting?

Nightenom:
> What is wrong with formatting?

it uses 16513156 spaces which causes each line to have one word which looks ugly?

also just thinking about it:
since this is not fixed map "from string to string", how does this work when:
a) chaining with future remaps
b) chaining with future remaps when intermediate blocks disappear/change that much so this method stops working correctly

marchermans:
> > What is wrong with formatting?
> 
> it uses 16513156 spaces which causes each line to have one word which looks ugly?
> 
> also just thinking about it:
> since this is not fixed map "from string to string", how does this work when:
> a) chaining with future remaps
> b) chaining with future remaps when intermediate blocks disappear/change that much so this method stops working correctly

With regards to the formatting:
image
That looks entirely normal to me.

With regards to the future remaps:
a) is just a new method,
b) is automatically taken care of, it won't remap anything it can't find.

Nightenom:
I don't know who are you but this looks 100x more readable to me :D (the colors are off cause of lightweight ide mode)
image

> b) is automatically taken care of, it won't remap anything it can't find.

This seems wrong to me, imagine the following case:
you have A block - this can be one of the dead blocks now
you replace A with B - that's what your commit consists of
then in the future block C comes
you create a mapping method from B to C because that the change you did, you really don't want to care about any older mappings like A to B (not to mention this effect will simply snowball later)

dfu is built in a way it takes care of such problems using levelled/versioned repairing
this design doesn't seem to me it can handle such problem

Raycoms:
LGTM :shipit:

marchermans:
> I don't know who are you but this looks 100x more readable to me :D (the colors are off cause of lightweight ide mode)
> image
> 
> > b) is automatically taken care of, it won't remap anything it can't find.
> 
> This seems wrong to me, imagine the following case:
> you have A block - this can be one of the dead blocks now
> you replace A with B - that's what your commit consists of
> then in the future block C comes
> you create a mapping method from B to C because that the change you did, you really don't want to care about any older mappings like A to B (not to mention this effect will simply snowball later)
> 
> dfu is built in a way it takes care of such problems using levelled/versioned repairing
> this design doesn't seem to me it can handle such problem

It does though.

The registry deals with this internally.

Nightenom:
I meant what if this becomes false? Then the chain cannot construct since it does not have an intermediate element?
image

marchermans:
Well obviously that is true, but since you can just change the list to try it against the more then one entry I don't see your point.

Yes both dfu and the registry can do this. But dfu can't update inventories you don't control, where this can.

So all in all this is far superior since it covers literally every case in existence in that little bit of code.

As an example: dfu would never be able to update C&B bits, because you would not even be able to detect it. But this can, and does straight out of the box.

Nightenom (approved):
Welp, while I get your opinion my point was basically that no one likes to care about old data fixing methods. A better solution from my point is to have a fixed RL to RL map, a) it does not have to take care about whether the end mapping exists or not (since someone checked that during testing), b) it can be chained nicely, c) you can reuse that everywhere (which might have no other use since this is tightened to registry)
side note: I'm not sure how you would handle TE to blockState (or vice versa) transformation with missing event mapping, might work with nongeneric event, but that's out of scope of this pr

PR #344, opened 2021-04-08
Expand View sizes when zero

ShadowProtocol:
Hopefully this does what Ray is looking for :smiley: 

- Allows widths and heights of 0 to instead fill out the rest of the parent container
    - Similar to using 100%, except
        - this way can be done automatically when the size is set, 
        - and only adjusts to the end of the container, as opposed to literally 100% of parent size but from its own position

Review please

### DIFF
--- a/src/main/java/com/ldtteam/blockout/Pane.java
+++ b/src/main/java/com/ldtteam/blockout/Pane.java
@@ -525,6 +525,9 @@ public class Pane extends AbstractGui
         if (parent != null)
         {
             parent.addChild(this);
+
+            // Allow views to expand zero-widths
+            setSize(width, height);
         }
     }
 
@@ -785,9 +788,6 @@ public class Pane extends AbstractGui
 
     /**
      * Handle onHover element, element must be visible.
-     *
-     * @param mx mouse x
-     * @param my mouse y
      */
     protected void handleHover()
     {
--- a/src/main/java/com/ldtteam/blockout/views/View.java
+++ b/src/main/java/com/ldtteam/blockout/views/View.java
@@ -35,6 +35,13 @@ public class View extends Pane
     public View(final PaneParams params)
     {
         super(params);
+
+        if (params.getParentView() != null) // might be null if created dynamically
+        {
+            if (width == 0) width = params.getParentView().width - x;
+            if (height == 0) height = params.getParentView().height - y;
+        }
+
         padding = params.getInteger("padding", padding);
     }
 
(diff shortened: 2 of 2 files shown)


Raycoms (approved):

someaddons (approved):

src/main/java/com/ldtteam/blockout/views/View.java
```diff
@@ -169,6 +176,33 @@ public void setWindow(final Window w)
...
+    {
+        super.setSize(w, h);
+
+        // Allow elements to have their size expanded when zero
+        View p = parent;
+        if (p == null) return;
+
+        if (width == 0)
+        {
+            while (p.width == 0 && p.parent != null)
```
Nightenom, line 190: This should never go higher than parent 
either parent is view - after this PR view cant have 0 size
or it's element and then it probably has size 0 for reason or because someone forgot to set it
so basically this is hiding bugs imho

src/main/java/com/ldtteam/blockout/views/View.java
```diff
@@ -169,6 +176,33 @@ public void setWindow(final Window w)
...
+            {
+                p = p.parent;
+            }
+            width = Math.max(0, p.width - x);
```
Nightenom, line 194: Also, this makes no sense, the more you go higher in the parent lookup the more this is wrong (p.width/height of anything but this.parent has not connectivity to this element, might be fixable if you count all the offsets during the while loop)

Nightenom (commented):
Meh, whatever, personally I don't like because "stop being lazy", but whatever, 2030 team member should not talk about this I guess

src/main/java/com/ldtteam/blockout/Pane.java
```diff
@@ -525,6 +525,9 @@ public void putInside(final View newParent)
         if (parent != null)
         {
             parent.addChild(this);
+
+            // Allow views to expand zero-widths
+            setSize(width, height);
```
Nightenom, line 530: if this instanceof View...

src/main/java/com/ldtteam/blockout/views/View.java
ShadowProtocol: Views can have 0 size if created dynamically with the default constructor - this is what I was accounting for here.
And parents are always Views, so the rest is fine.

src/main/java/com/ldtteam/blockout/views/View.java
ShadowProtocol: Now that is interesting, I think the offset of grandparents simply didn't occur to me.This will indeed create a bug, though a few things would have to be specified in certain ways to produce it. Anyways yeah I'll account for offset in a future PR. Good catch.

src/main/java/com/ldtteam/blockout/Pane.java
ShadowProtocol: Due to polymorphism this doesn't really matter much. Though arguably that may have been more maintainable.

ShadowProtocol (commented):
I did this upon request, just thought I'd put that out there.

I disagree with the 'stop being lazy' because as per the description it means a slightly different thing to other values like 100%, it allows for better dynamic creation, and UI testing in minecraft is already inconvenient, why make everyone fiddle with numbers on top of it?

src/main/java/com/ldtteam/blockout/views/View.java
Nightenom: Oh, I overlooked that you made changes to params ctor, not the default one. Still, I would choose new method (eg. setSizeFromParent) and leave the setSize method unchanged over this iterative solution :D

PR #357, opened 2021-04-17
Slightly Smarter Workaround For J8 vs J9+ problems

gattsuru:
- Comments out the IBlockCollection.create default method, adds implementing @override exact copies into BrickType and CactusCollection.

Yes, this is dumb.

Tested against AdoptOpenJDK JREs 11.0.10.9, 12.0.1.12, 15.0.2.7, as well as Oracle 8u271 and 8u58.  AdoptOpenJDK 16.0.0.36 _doesn't_ work, but I don't think Forge likes it in general yet.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/blocks/CactusCollection.java
+++ b/src/main/java/com/ldtteam/structurize/blocks/CactusCollection.java
@@ -10,8 +10,13 @@ import net.minecraft.block.Block;
 import net.minecraft.block.Blocks;
 import net.minecraft.data.IFinishedRecipe;
 import net.minecraft.data.ShapelessRecipeBuilder;
+import net.minecraft.item.BlockItem;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemGroup;
 import net.minecraftforge.fml.RegistryObject;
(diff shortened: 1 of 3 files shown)


Nightenom:
I would rather go for anonymous supplier class which should do the same work, but not tested

gattsuru:
That does seem a lot more clean.  Tested against the same batch of JREs, seems to be working fine in all of them.

src/main/java/com/ldtteam/structurize/blocks/CactusCollection.java
```diff
@@ -10,8 +10,13 @@
 import net.minecraft.block.Blocks;
 import net.minecraft.data.IFinishedRecipe;
 import net.minecraft.data.ShapelessRecipeBuilder;
+import net.minecraft.item.BlockItem;
+import net.minecraft.item.Item;
```
ravenbuilder934: Imports?

gattsuru: Thanks.

Raycoms:
:shipit:

marchermans (approved):

Nightenom (approved):

PR #362, opened 2021-04-24
Circular algo

Raycoms:
- Create a circular (outside-in) algorithm for block placement within buildings.
- Add config option to enable it (default disabled)
- This requires a little bit of work on the MineColonies side, to store the version with a buildorder such that the structure stage is reset if the version changes.
- This is on purpose public static such that this can't change during runtime.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
+++ b/src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
@@ -1,6 +1,5 @@
 package com.ldtteam.structurize.config;
 
-import com.google.gson.JsonArray;
 import net.minecraftforge.common.ForgeConfigSpec;
 
 /**
--- a/src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
+++ b/src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
@@ -1,5 +1,6 @@
 package com.ldtteam.structurize.config;
 
+import com.google.gson.JsonArray;
 import net.minecraftforge.common.ForgeConfigSpec;
 
 /**
@@ -44,6 +45,11 @@ public class ServerConfiguration extends AbstractConfiguration
     public final ForgeConfigSpec.IntValue schematicBlockLimit;
 
     /**
+     * Iteration version this structurize version is using.
+     */
+    public final ForgeConfigSpec.IntValue iterationVersion;
+
+    /**
      * Builds server configuration.
      *
      * @param builder config builder
@@ -59,6 +65,7 @@ public class ServerConfiguration extends AbstractConfiguration
         maxCachedSchematics = defineInteger(builder, "maxCachedSchematics", 100, 0, 100000);
         maxBlocksChecked = defineInteger(builder, "maxBlocksChecked", 1000, 0, 100000);
         schematicBlockLimit = defineInteger(builder, "schematicBlockLimit", 100000, 1000, 1000000);
+        iterationVersion = defineInteger(builder, "iterationVersion", 1, 1, 2);
 
         finishCategory(builder);
     }
--- a/src/main/java/com/ldtteam/structurize/placement/BlueprintIterator.java
+++ b/src/main/java/com/ldtteam/structurize/placement/BlueprintIterator.java
@@ -88,7 +88,7 @@ public class BlueprintIterator
      */
     public Result iterate(final boolean up)
     {
-        if (ITERATION_VERSION < 1)
+        if (ITERATION_VERSION <= 1)
         {
             return up ? increment() : decrement();
         }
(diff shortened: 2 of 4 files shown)


src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
```diff
@@ -59,6 +64,7 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
         maxCachedSchematics = defineInteger(builder, "maxCachedSchematics", 100, 0, 100000);
         maxBlocksChecked = defineInteger(builder, "maxBlocksChecked", 1000, 0, 100000);
         schematicBlockLimit = defineInteger(builder, "schematicBlockLimit", 100000, 1000, 1000000);
+        iterationVersion = defineInteger(builder, "iterationVersion", 1, 1, 2);
```
uecasm, line 67: This isn't especially user-friendly.  Though I suppose it could be explained in comments.

src/main/java/com/ldtteam/structurize/placement/BlueprintIterator.java
```diff
@@ -47,6 +48,19 @@
...
+     */
+    private static int ITERATION_VERSION = Structurize.getConfig().getServer().iterationVersion.get();
+
+    /**
+     * Tracking current state.
+     */
+    private int min_x;
+    private int max_x;
+    private int min_z;
+    private int max_z;
```
uecasm, line 62: Perhaps consider using `MutableBoundingBox`?  Though that has Y as well.

Raycoms: Yeah, that's why I didn't use any other.

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
Raycoms: Actually, I did set that on purpose for the time being. I want this added initially that people that are open for experiments can test and not sth people in general should be playing around with on/off.

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -11,7 +11,6 @@
 import net.minecraft.block.material.Material;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityType;
-import net.minecraft.entity.LivingEntity;
```
ravenbuilder934, line 14: *yes*

src/main/java/com/ldtteam/structurize/placement/BlueprintIterator.java
someaddons: the position we expose should be immutable though

Raycoms: He only means the tracking variables here

src/main/java/com/ldtteam/structurize/placement/BlueprintIterator.java
```diff
@@ -47,6 +48,19 @@
      */
     private boolean isRemoving;
 
+    /**
+     * Iteration version (1 = original, 2 = outside in).
+     */
+    private static int ITERATION_VERSION = Structurize.getConfig().getServer().iterationVersion.get();
```
marchermans, line 54: This does not work!

Server config is not available on the load of this class as such the static value will most likely differ between client and server!

marchermans (changes requested):

src/main/java/com/ldtteam/structurize/placement/BlueprintIterator.java
Raycoms: This only runs on the server.

Raycoms: Or do you mean singleplayer/smp?

uecasm: I think the point is that in theory, static constructors run at mod init, which is before the config files are read.  (In practice, static constructors run whenever the class is actually loaded, which is more fragile / less predictable -- and for this particular class, probably does work by happenstance.)

marchermans: @uecasm Is correct.
Additionally this completely breaks in SMP/SIngleplayer.

With my PR to forge you can just query it all the time, there is virtually no performance benefit to doing it this way.

Raycoms: I don't do this for performance, I don't want it changed during runtime.

marchermans: Then don't make it configurable.

The forge config system was designed around mutability of these values.
Especially the Server configuration has this in mind since it is synchronized from the client.

What is the exact reason you don't want this to change during runtime? I mean I can see why you don't want it to change during an iteration/buildprocess. But when no build is going on, does it matter that the value changes?

PR #371, opened 2021-05-04
Need to call this

Raycoms:
- Fix banners calling this event
-
-

Review please


Mekle001 (approved):

Nightenom (commented):
This method should be called by all placers afaik. Not sure now

Raycoms:
> This method should be called by all placers afaik. Not sure now

Yeah, I'm not 100% sure. Either we can handle it "on success" on a higher level, or if we should put it somehow in.
But I'm not 100% sure about side-effects like performance when pasting. I only know we need this specifically for the banner atm (where not having it has a negative side-effect). So for the time being I only put it here

PR #378, opened 2021-05-24
Add input handler to textfields

Raycoms:
- Adds input event handling
-
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/blockout/controls/TextField.java
+++ b/src/main/java/com/ldtteam/blockout/controls/TextField.java
@@ -573,17 +573,9 @@ public class TextField extends Pane
     {
         InputHandler delegatedHandler = handler;
 
-        if (delegatedHandler == null)
+        if (delegatedHandler == null && getWindow() instanceof InputHandler)
         {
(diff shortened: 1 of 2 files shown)


marchermans (dismissed):

src/main/java/com/ldtteam/blockout/controls/TextField.java
```diff
@@ -626,6 +663,8 @@ public void deleteFromCursor(final int count)
                 this.moveCursorBy(count);
             }
         }
+
+        triggerHandler();
```
Nightenom, line 659: Might probably be double call because of line 602

Nightenom (dismissed):

src/main/java/com/ldtteam/blockout/controls/TextField.java
Raycoms: Should be if/else there too, I tried it out and it would only trigger once for me

Nightenom (approved):
mark for myself, add few ticks delay to mitigate fasttypers update each tick

PR #384, opened 2021-06-03
Fix items not being put into chests.

### LINKED ISSUE #383: Chests are placed empty
### Prerequisites 

- [x] I am running the latest **alpha** version of Structurize for my Minecraft version.
- [x] I checked the MineColonies/Structurize [wiki](https://wiki.minecolonies.ldtteam.com/) and made sure my issue is not covered there.
- [x] I made sure that **this issue is not a duplicate of any existing issue.**

### Context 

- Minecraft Version: 1.16.5
- Structurize Version: 0.13.175
- MineColonies Version (if related bug): 0.14.197
- Related Mods and their Versions:

## Expected behavior

Building or pasting a schematic should include items in chests.

## Actual behavior

The chests appear but they are always empty.  The blueprint file itself does contain the item data, so it's the build that's failing rather than the scan.

## Steps to reproduce the issue

1. Try to build the `asiansupplycamp`, either with a normal supply camp item or with either of the build tool paste buttons.
2. Note that all the chests are empty, but there are items in the racks and barrels.  (But there is supposed to be a build tool and TH block in one of the chests, which you can see in the blueprint data and on the schematic server.)
3. Also try building this schematic: [chest-test.zip](https://github.com/ldtteam/Structurize/files/6591560/chest-test.zip); it should contain 64 redstone blocks.

### Logs

There is nothing relevant in logs.  This was reproduced in a dev environment with no other mods.

### Notes

---
## Viewers

uecasm:
Closes #383

- Fixes chests being placed empty.

Review please

Removing that redundant bit of code (the return value is ignored, the only thing it does is to throw an exception and return because it was given the wrong world) fixes everything.  The chest is populated as expected.  Double chests also work as expected and don't duplicate anything, no matter what rotation is used.  The only thing that continues to not work is mirroring a double chest -- but even then it has the right items, it just doesn't connect properly (which is not new).

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
+++ b/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
@@ -718,17 +718,6 @@ public final class PlacementHandlers
                 return ActionProcessingResult.DENY;
             }
 
-            try
-            {
-                // Try detecting inventory content.
-                ItemStackUtils.getItemStacksOfTileEntity(tileEntityData, world, pos);
-            }
-            catch (final Exception ex)
-            {
-                // If we can't load the inventory content of the TE, return early, don't fill TE data.
-                return ActionProcessingResult.SUCCESS;
-            }
-
             if (tileEntityData != null)
             {
                 handleTileEntityPlacement(tileEntityData, world, pos);
--- a/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
+++ b/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
@@ -10,7 +10,6 @@ import net.minecraft.block.*;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.item.ItemEntity;
-import net.minecraft.inventory.ItemStackHelper;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.Items;
 import net.minecraft.nbt.CompoundNBT;
@@ -19,7 +18,6 @@ import net.minecraft.state.properties.BlockStateProperties;
 import net.minecraft.state.properties.DoubleBlockHalf;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.Direction;
-import net.minecraft.util.NonNullList;
 import net.minecraft.util.math.AxisAlignedBB;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
@@ -56,7 +54,6 @@ public final class PlacementHandlers
         handlers.add(new SpecialBlockPlacementAttemptHandler());
         handlers.add(new FlowerPotPlacementHandler());
         handlers.add(new StairBlockPlacementHandler());
-        handlers.add(new ChestPlacementHandler());
         handlers.add(new ContainerPlacementHandler());
         handlers.add(new FallingBlockPlacementHandler());
         handlers.add(new BannerPlacementHandler());
@@ -224,7 +221,7 @@ public final class PlacementHandlers
           @Nullable final CompoundNBT tileEntityData,
           final boolean complete)
         {
-            final List<ItemStack> itemList = new ArrayList<>(getItemsFromTileEntity(tileEntityData, world, pos));
+            final List<ItemStack> itemList = new ArrayList<>(getItemsFromTileEntity(tileEntityData, blockState));
             itemList.add(BlockUtils.getItemStackFromBlockState(blockState));
             itemList.removeIf(ItemStackUtils::isEmpty);
             if (!world.getBlockState(pos.down()).getMaterial().isSolid())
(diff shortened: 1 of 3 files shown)


Raycoms (changes requested):
No, this is here on purpose, if you remove this, players can infinitely dupe items having the builder build chests with content in them.

uecasm:
Why?  The builder should demand the items that are in the inventory.  But that would have to happen elsewhere, because this bit of code can't possibly report that anyway.

uecasm:
`getRequiredItems` is an entirely separate code path.

Raycoms:
> Why? The builder should demand the items that are in the inventory. But that would have to happen elsewhere, because this bit of code can't possibly report that anyway.

The problem is this:
- In the same handler a bit below this code there is a method to query the items. Since chests suck, the items return empty.
- As such the builder won't request any items.
- As such the builder will reach this stage.
- As such, we have to detect in this stage if the items could be queried or not, and if not return an empty inventory.

uecasm:
Perhaps then this should only happen if it's a builder operation rather than a creative paste?  Because I can see this being a problem for raid ship chests too as it stands...

Raycoms:
> Perhaps then this should only happen if it's a builder operation rather than a creative paste? Because I can see this being a problem for raid ship chests too as it stands...

The solution is simply to make a propper chest handler that is only called for those vanilla containers and that AT the inventory or read the inventory from NBT such that no crash happens.

uecasm:
Ok, there, that should work.  I haven't tested this with a builder as yet (but I might try in the morning if nobody beats me to it) but I'm confident it will do the job.

Raycoms (dismissed):

Nightenom (dismissed):

uecasm:
FWIW, this `loadAllItems` call appears to be used for all of the following in vanilla:
* multiple kinds of furnaces
* barrels
* brewing stands
* campfires
* chests
* dispensers
* hoppers
* shulker boxes

The "since chests suck" code appears to be entirely added by Forge, incidentally (not present in vanilla), and as far as I can see does only affect chests and not any of the other above blocks.  Having said that, it's ultimately still a bug in Structurize (not in Forge) where `getItemStacksOfTileEntity` is passing the wrong World and BlockState to `readTileEntity` during placement (it's passing the target world and block therein, not the blueprint); it's simply a happy accident that most TEs don't seem to care about that.  (This PR is still an improvement, but it would be good to fix the rest of it later.)

Raycoms:
> * dispensers

The thing is, that it's impossible to handle double chests otherwise. Even if we create a simulated world just for the placement of chests. We would still get duplicate items from it. As that, on purpose I don't give it the world such that it cannot get the chest. I rather have an empty inventory and place an empty inventory than allowing to dupe items.

Nightenom:
Solution to this would be proper multiblock handler imho
With the current state single block nbt parsing is the best thing I see

Raycoms:
In the "canHandle" method, we should be able to check if the tileentitydata contains regular inventory data, no?

Nightenom:
> In the "canHandle" method, we should be able to check if the tileentitydata contains regular inventory data, no?

Yeah, surely, but remember that this check should be as fast as possible

Raycoms:
if container && tenbt.containsx should match most use cases

uecasm:
Sure.  Want me to make that change?  Even though with only vanilla blocks it wouldn't actually be needed?  (The other blocks with compatible NBT data aren't multiblocks.)

Raycoms:
Sure

Raycoms (approved):
I like this

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -689,14 +689,14 @@ public ActionProcessingResult handle(
...
-            final List<ItemStack> itemList = new ArrayList<>(getItemsFromTileEntity(tileEntityData, world, pos));
+            final List<ItemStack> itemList = new ArrayList<>(getItemsFromTileEntity(tileEntityData, blockState));
             itemList.add(BlockUtils.getItemStackFromBlockState(blockState));
             itemList.removeIf(ItemStackUtils::isEmpty);
             return itemList;
         }
     }
 
-    public static class ChestPlacementHandler implements IPlacementHandler
+    public static class ContainerPlacementHandler implements IPlacementHandler
```
Nightenom, line 699: Please undo the class rename, as it doesn't seem necessary

Nightenom (commented):
Minor nitpick

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
uecasm: It handles ContainerBlock, and not just ChestBlock.  The rename seems entirely necessary.

Raycoms: I agree

Nightenom (approved):

PR #385, opened 2021-06-03
Populates the `required_mods` list properly.

uecasm:
- Correctly populates the `required_mods` list in the blueprint data when scanning.

Review please

----
This fixes a bug I noted in ldtteam/minecolonies#7190.  Nothing appears to actually *use* this list at present (much like the `architects` list) but it seems useful to have it correct such that in a future version of structurize it could either automatically hide or just display an error message if you tried to build a schematic that requires blocks from unloaded mods.  Or as part of a schematic lint checker to verify that stray blocks didn't make it into an official style (or to relax that rule if such styles could be automatically hidden when not usable).

### DIFF
--- a/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java
+++ b/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java
@@ -23,6 +23,8 @@ import java.io.OutputStream;
 import java.util.*;
 import java.util.function.Function;
 
+import static com.ldtteam.structurize.api.util.constant.Constants.MOD_ID;
+
 /**
  * @see <a href="http://dark-roleplay.net/other/blueprint_format.php">Blueprint V1 Specification</a>
  * @since 0.1.0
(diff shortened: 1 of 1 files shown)


Raycoms (approved):

uecasm:
Note that while I tried to follow what looked like the intent of the original code (w.r.t. clearing the block if the mod isn't loaded), that part seems kinda pointless, as it shouldn't be possible to have a block from an unloaded mod in the world -- although it **is** possible to have a registered block with a namespace not matching any loaded mod (though that's generally considered a bug in whichever mod registered it).

Also I stand corrected, the `required_mods` list does get used on load to check for missing mods as expected (I was searching for uses of the Blueprint method, but this happens before it gets that far).  So this PR should fix that actually working, no further future change needed. 😉

Raycoms:
I think it was supposed to ignore th eblocks in th eblueprint when putting them in the world if th emod isn't loaded, and not the other way around.

uecasm:
Well, that will happen anyway; since it can't construct a BlockState from an unregistered block it will just end up as AIR. 🤷‍♂️ Leaving this in isn't going to hurt anything, at least.

marchermans:
You are both correct

The problem is that the missing block can not become air, if it never was part of the world in the first place as such @Raycoms is correct.

If you however tried to say have a mod removed and then load a schematic with it, then @uecasm is correct, FML will create dummy air blocks, dummy items, dummy tile entities etc in the registry for those missing objects and structurize can happily use them. But only after the mod has existed at least once in the world and it's registry objects are given a save id.

So the long and short of it is: We need skip any registry objects which does not exist in the namespace. So check for blocks and tes during the placement phase. Check for items during the population phase, check for entities during their phase, etc.

@uecasm it would be good if you could run through these code paths and validate that what I just described is actually also validated.

src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java
```diff
@@ -70,20 +72,17 @@ public static Blueprint createBlueprint(World world,
...
-                    {
-                        requiredMods.add(modName);
-                    }
-                }
-                else if (!ModList.get().getModContainerById(modName).isPresent())
+                if (!ModList.get().getModContainerById(modName).isPresent())
                 {
                     structure[y][z][x] = (short) pallete.indexOf(Blocks.AIR.getDefaultState());
                     continue;
                 }
```
uecasm, line 81: To clarify, I'm saying that this part of the mostly-existing scan code is pointless, but also harmless (unless some mod registers blocks in an unrelated namespace).  So it could be removed if people feel strongly enough about it but I was inclined to leave it for backwards consistency.  The rest of the discussion wasn't really related to any code in this PR.

Raycoms: Well, this is blueprint creation, so indeed it would be weird trying to scan a black that's not present (as far as I understood).

Nightenom (approved):

PR #392, opened 2021-06-13
Tabs (and other minor blockout things)

ShadowProtocol:
- Adds a formal tab system to Blockout for easy creation of switch-like tabbed views
    - Designed to fulfil the current use case in mcol, plus a couple more planned uses 
    - Both dynamic and declarable forms, with various options including vertical or horizontal tab lists
    - Has simpler XML and other subtle advantages over the system currently used in mcol.
- Rectifies a dormant bug where the widths and heights were switched in button image setters
- Added a shorthand for resources, especially textures
    - For example, can collapse  ``structurize:textures/gui/buildtool/confirm.png`` down to ``@:buildtool/confirm`` in XML

Review please


Raycoms (dismissed):
Looks good

src/main/java/com/ldtteam/blockout/views/Tab.java
```diff
@@ -0,0 +1,71 @@
...
+     * NOTE: this will not add it to a tab set, for that use {@link TabSet#addTab(Tab)}
+     * @param title the label for the tooltip
+     * @param icon a resource string to the icon relative to the gui textures folder
+     * @param content each element to be included in this view
+     */
+    public Tab(String title, String icon, Pane... content)
+    {
+        this.label = new StringTextComponent(title);
+        this.icon = Parsers.RESOURCE("textures/gui/", "png").apply(icon);
+        for (Pane p : content) addChild(p);
```
Nightenom: meh, remove, you can use addChild manually

src/main/java/com/ldtteam/blockout/views/Tab.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public Tab(PaneParams params)
+    {
+        super(params);
+
+        setPosition(0, 0);
+        setSize(params.getParentWidth(), params.getParentHeight());
+
+        this.icon = params.getTexture("icon", $ -> {});
```
Nightenom: you should create function alias if loader is not needed (or at least accept null loader)

src/main/java/com/ldtteam/blockout/views/Tab.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    /**
+     * Constructs a text item for this tab view
+     * @return the text pane
+     */
+    public Text buildLabel()
+    {
+        ITextComponent component = label == null ? new StringTextComponent("") : label;
```
Nightenom:
```suggestion
        ITextComponent component = label == null ? StringTextComponent.EMPTY : label;
```

src/main/java/com/ldtteam/blockout/views/Tab.java
```diff
@@ -0,0 +1,71 @@
...
+    protected IFormattableTextComponent label;
+
+    /**
+     * Constructs a new tab from the given information.
+     * NOTE: this will not add it to a tab set, for that use {@link TabSet#addTab(Tab)}
+     * @param title the label for the tooltip
+     * @param icon a resource string to the icon relative to the gui textures folder
+     * @param content each element to be included in this view
+     */
+    public Tab(String title, String icon, Pane... content)
```
Nightenom: suggestion: might want to hide that and use ctor directly in TabSet#addChild()

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+
+import java.util.Random;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Similar to a switch view, the tab set renders the selected view,
+ * allowing for tabs to be placed for user's selection
+ */
+public class TabSet extends View
```
Nightenom: suggestion: we should keep using the view naming ie. TabView

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+          .peek(child -> child.padding = childPadding)
+          .forEach(ButtonTab::new);
+
+        setSelectedTab(0);
+    }
+
+    @Override
+    protected boolean childIsVisible(final Pane child)
+    {
+        // Draw the tabs despite being potentially out of the tab set view
```
Nightenom: This should be elemently-inlined (not sure how to say properly :D ) just like scrollbar in ScrollingView - this is overall hard to design.
We should propably invent a better mechanism for auto elements like this and scrollbar: something like the element is in the xml file with fixed id which allows the xml writer to define it's size and other arguments - this obviously add unwanted xml which you have to remember to put in.
Feel free to come with better solution

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+        super(params);
+        vertical = params.getBoolean("vertical", false);
+        overlap = params.getInteger("overlap", 0);
+
+        tabWidth = params.getInteger("tabwidth", 100);
+        tabHeight = params.getInteger("tabheight", 20);
+        offset = params.getInteger("taboffset", 0);
+
+        childPadding = padding;
+        padding = 0;
```
Nightenom: huh, probably connected to comment below, but why don't you just use padding of this element?

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+    public class ButtonTab extends View
+    {
+        protected Tab   toSelect;
+        protected Image background = new Image();
+        protected Image icon = new Image();
+
+        public ButtonTab(Tab tab)
+        {
+            super();
+            int index = tabs.children.size();
```
Nightenom: way hacky to me, parent (TabSet) should pass this as param

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+ */
+public class TabSet extends View
+{
+    private static final Random RANDOM = new Random();
+
+    protected boolean vertical;
+    protected int spacing;
+    protected int overlap;
+    protected int offset;
+    protected int tabWidth;
```
Nightenom: this seems related to buttonTab, not just tab

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+            ? mx >= x + overlap - tabWidth && mx < (x + width) && my >= y && my < (y + height)
+            : mx >= x && mx < (x + width) && my >= y + overlap - tabHeight && my < (y + height);
+    }
+
+    /**
+     * The pane for the physical representation of the tab itself,
+     * i.e. the button you click to change content.
+     * Has a background and a potential icon or text otherwise
+     */
+    public class ButtonTab extends View
```
Nightenom: suggestion: TabButton?

src/main/java/com/ldtteam/blockout/views/TabSet.java
```diff
@@ -0,0 +1,271 @@
...
+        public void drawSelf(final MatrixStack ms, final double mx, final double my)
+        {
+            if (toSelect.equals(selectedTab)) return;
+            super.drawSelf(ms, mx, my);
+        }
+
+        @Override
+        public void drawSelfLast(final MatrixStack ms, final double mx, final double my)
+        {
+            if (toSelect.equals(selectedTab)) super.drawSelf(ms, mx, my);
```
Nightenom: This is a no, this pass should used for overlays which needs to sit on top of screen, not for delayed rendering. Use small z translate in drawSelf instead, if you wanted to achieve this being higher than base element - I would like to know why this even exists

src/main/java/com/ldtteam/blockout/views/TabSet.java
Nightenom: Another thing, this should extend Button, not View, because of disabled/hover images etc. so one can use disabled image for tab button being selected or not (propably use Pane.enabled to disable when selected)

Nightenom (changes requested):
I'm not happy with the texture shorcuts:
- the [todo comment](https://github.com/ldtteam/Structurize/pull/392/files#diff-a0a91f1e8392fb67679db7f500d751ba6e2ff7c77bac073b07cc6412b3852f19R196) should be fixed before merging
- I would go with "if namespace not present use parent" rather than "@:" 
- 99% I would not remove the suffix (.png) from path
- you should try existence of original RL, then with textures/gui if original not found

Generally I would like to have common RL test which would test RLs during xml parsing, if they are not found use minecraft missing texture and log error

Formatting everywhere

The ButtonTab really should be button, the width of that button should be automatical with regards to either icon or text, text should be rotated when left-sided tabs are used (probably want to use param for that, sth like rotation angle)

Sorry for me being hateful, I really like the idea of tabbed view being directly in blockout, but I can't get rid of the feeling that this could be way better in terms of code

src/main/java/com/ldtteam/blockout/views/TabSet.java
ShadowProtocol: They sure are related to tabs. Can prefix these to make it clearer, but I'm not going to write XML attributes in every tab that are all the same if I can just write them in the tabview instead.

src/main/java/com/ldtteam/blockout/views/TabSet.java
ShadowProtocol: Planned use case. Some direct children in the TabView that aren't tabs may wish to use zero, such as images for backgrounds. This way also ensures consistency across tabs without having to write it all in XML.

src/main/java/com/ldtteam/blockout/views/Tab.java
ShadowProtocol: Not quite sure what you mean but this will be used for dynamic element creation by planned use cases. It's fine as it is, and I removed the vararg.

someaddons (approved):

Raycoms:
@Nightenom ? is this ready?

Raycoms (approved):

ShadowProtocol:
It's kinda ready, but not to a great standard. It works and is fairly robust, but its not the cleanest code I've ever made and uses two minor hacks.
Night and I discussed it a bit some weeks back, he was going to try something from what I'd gathered, regarding one of those hacky parts with the positioning of the tab rail. I've turned my attention to the loader system since.

Raycoms:
1.16 is EOL

Nightenom:
Tru, but this is still not done, yet planned to add

Raycoms:
Then make it an issue, I don't think we need the PR? As you were unhappy anyway and it would be on another repo.

Nightenom:
Why dupe things, when single ref is enough..
I'm unhappy with elements with fixed childs, which need much more work than expected..

PR #397, opened 2021-06-21
Bump optifine compat

Nightenom:
- bump optifine compat to work properly in g7/g8 versions

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
+++ b/src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
@@ -263,13 +263,7 @@ public class BlueprintRenderer implements AutoCloseable
         renderBufferSource.finish(Atlases.getShulkerBoxType());
         renderBufferSource.finish(Atlases.getSignType());
         renderBufferSource.finish(Atlases.getChestType());
-        if(OptifineCompat.getInstance().isOptifineEnabled())
-        {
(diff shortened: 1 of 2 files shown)


Raycoms:
Oh cmon, fix the whitespacing of your IDE xD Or git setup

Nightenom:
> Oh cmon, fix the whitespacing of your IDE xD Or git setup

Did change the config now, haven't touched it before tho :D

marchermans (changes requested):
Fix your whitespacing.
This is not acceptable in the current state.

Nightenom:
> Fix your whitespacing.
> This is not acceptable in the current state.

You know you can hide whitespace changes in Github files page? and also I've said we should add .gitattribute with autocrlf to make sure git doesn't explode for everyone, you've said no - really not sure why though

if you really want me to change this commit then put reaction on this (im not now able to do so, will do tomorrow if needed)

uecasm:
`autocrlf` is evil.  Although I do recommend setting `.gitattributes` with `* -text` to force all clients to **not** do CRLF transformations, rather than relying on people on Windows to use the correct settings instead of the defaults.  Though that might be a pain in the butt initially for those people using the incorrect defaults.

Raycoms (approved):

marchermans (approved):

PR #398, opened 2021-06-22
Circ alg

Raycoms:
Closes #
Closes #
Closes #

- Setup system to have different iterator modes and allow other mods to extend this with their own implementations.
- Add two new modes: a) Circular and b) Circular with varying heights.
-

Review please


src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
```diff
@@ -59,6 +62,7 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
         maxCachedSchematics = defineInteger(builder, "maxCachedSchematics", 100, 0, 100000);
         maxBlocksChecked = defineInteger(builder, "maxBlocksChecked", 1000, 0, 100000);
         schematicBlockLimit = defineInteger(builder, "schematicBlockLimit", 100000, 1000, 1000000);
+        iteratorType = defineString(builder, "iteratorType", "default");
```
Nightenom, line 65: Can this have validator, something like `str.lowercase() in possibleValues`?

Raycoms: Where?

uecasm: Perhaps it should be an EnumValue instead?  Although I guess `registerIterator` implies that it's intended that other mods could add custom iterators.

src/api/java/com/ldtteam/structurize/api/util/BlockPosUtil.java
```diff
@@ -211,4 +212,124 @@ public static boolean isInbetween(final BlockPos inside, final BlockPos corner1,
...
+
+    /**
+     * Gets the next position to use in a circular-creeping inward snake manner. Its iterating the rectangle of corners in clockwise rotation
+     *
+     * @param start   current positions
+     * @param corner1 rectangle top left corner
+     * @param corner2 rectangle bottom left corner
+     * @return next position
+     */
+    public static BlockPos getNextPosInCircleFrom(final BlockPos start, final BlockPos corner1, final BlockPos corner2, final int ringHeight)
```
uecasm, line 225: Missing javadoc for `ringHeight`.  Also presumably `corner2` should be bottom right?

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
Raycoms: Yes, the main reason it's not an Enum is that a) Other mods can add iterators and b) We can add for testing purposes on the MineColonies side iterators easily without having to recompile structurize etc.

src/api/java/com/ldtteam/structurize/api/util/BlockPosUtil.java
Raycoms: @someaddons tststs

src/main/java/com/ldtteam/structurize/placement/StructureIterators.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param id      the unique id.
+     * @param handler the handler.
+     * @return the instance of the iterator.
+     */
+    public static AbstractBlueprintIterator getIterator(final String id, final IStructureHandler handler)
+    {
+        final Function<IStructureHandler, AbstractBlueprintIterator> iterator = iteratorProducers.get(id);
+        if (iterator == null)
+        {
+            Log.getLogger().warn("Could not find iterator for value:" + id + " using default instead!");
```
Nightenom, line 54: yep, afaik the config value has some kind of validator in it, also I would rather crash during mod load if the value is invalid

Raycoms: Crashing is always a bad idea, invalid value = default iterator + warning is good enough.

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
Nightenom: we can still emulate enum in the way vanilla does it with eg. RenderType

Raycoms: I don't really think that is worth it.

src/main/java/com/ldtteam/structurize/placement/StructureIterators.java
Nightenom: yeah, but if you put wrong arguments to cli command then it crash, config is just neat way to have many program args
the problem is that log warning is usually not noticed by end-users, so at least send chat message on world join

Raycoms: I think printing an exception here might actually be an idea. People do read and report those usually.

PR #406, opened 2021-07-03
Generalize Hut Checks

Justen-N:
- Implements a more generalized check for hut blocks 
- Instead of limiting check to the Minecolonies namespace, the check is now performed against each origin mod's namespace. 
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/management/StructureName.java
+++ b/src/main/java/com/ldtteam/structurize/management/StructureName.java
@@ -1,6 +1,5 @@
 package com.ldtteam.structurize.management;
 
-import com.ldtteam.structurize.blocks.interfaces.IAnchorBlock;
 import com.ldtteam.structurize.util.LanguageHandler;
 import com.ldtteam.structurize.api.util.constant.Constants;
 import com.ldtteam.structurize.util.StructureLoadingUtils;
@@ -22,6 +21,8 @@ import java.util.regex.Pattern;
  */
 public class StructureName
 {
+    public static List<String> HUTS = new ArrayList<>();
+
     private static final Pattern levelPattern              = Pattern.compile("[^0-9]+([0-9]+)$");
     private static final String  LOCALIZED_SCHEMATIC_LEVEL = "com.ldtteam.structurize.gui.buildtool.hut.level";
     /**
@@ -111,7 +112,7 @@ public class StructureName
             //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
             for (String origin : StructureLoadingUtils.getOriginMods().keySet())
             {
-                if (IAnchorBlock.class.isAssignableFrom(ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)).getClass()))
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
                 {
                     section = hut;
                     break;
--- a/src/main/java/com/ldtteam/structurize/management/StructureName.java
+++ b/src/main/java/com/ldtteam/structurize/management/StructureName.java
@@ -110,9 +110,9 @@ public class StructureName
             hut = schematic.split("\\d+")[0];
             section = Structures.SCHEMATICS_PREFIX;
             //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
-            for (final Map.Entry<String, ModFileInfo> origin : StructureLoadingUtils.getOriginMods().entrySet())
+            for (String origin : StructureLoadingUtils.getOriginMods().keySet())
             {
-                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin.getKey(), "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
                 {
                     section = hut;
                     break;
--- a/src/main/java/com/ldtteam/structurize/management/StructureName.java
+++ b/src/main/java/com/ldtteam/structurize/management/StructureName.java
@@ -2,13 +2,16 @@ package com.ldtteam.structurize.management;
 
 import com.ldtteam.structurize.util.LanguageHandler;
 import com.ldtteam.structurize.api.util.constant.Constants;
+import com.ldtteam.structurize.util.StructureLoadingUtils;
 import net.minecraft.block.Blocks;
 import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.fml.loading.moddiscovery.ModFileInfo;
 import net.minecraftforge.registries.ForgeRegistries;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -106,13 +109,14 @@ public class StructureName
         {
             hut = schematic.split("\\d+")[0];
             section = Structures.SCHEMATICS_PREFIX;
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
-            {
-                section = hut;
-            }
-            else
-            {
-                hut = "";
+            //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
+            for (final Map.Entry<String, ModFileInfo> origin : StructureLoadingUtils.getOriginMods().entrySet()) {
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin.getKey(), "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut)) {
+                    section = hut;
+                    break;
+                } else {
+                    hut = "";
+                }
             }
         }
     }
--- a/src/main/java/com/ldtteam/structurize/management/StructureName.java
+++ b/src/main/java/com/ldtteam/structurize/management/StructureName.java
@@ -21,7 +21,6 @@ import java.util.regex.Pattern;
  */
 public class StructureName
 {
-    public static List<String> HUTS = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -106,13 +109,14 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
...
             section = Structures.SCHEMATICS_PREFIX;
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
-            {
-                section = hut;
-            }
-            else
-            {
-                hut = "";
+            //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
+            for (final Map.Entry<String, ModFileInfo> origin : StructureLoadingUtils.getOriginMods().entrySet()) {
```
Raycoms: Seems okay, just the formatting is totally off.

Justen-N: What about the formatting needs fixed? Other than my somewhat unnecessary brackets? (former C developer, I'm a bracket addict.

Justen-N: I think I see it now. I'm dumb

Justen-N: Wonder where the setting in Intellij that is f-ing with that is...

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -106,13 +109,18 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
         {
             hut = schematic.split("\\d+")[0];
             section = Structures.SCHEMATICS_PREFIX;
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
+            //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
+            for (final Map.Entry<String, ModFileInfo> origin : StructureLoadingUtils.getOriginMods().entrySet())
```
Nightenom: You should be fine with key view (you don't seem to use value from entry set)

Nightenom (dismissed):

src/main/java/com/ldtteam/structurize/management/StructureName.java
Justen-N: You are absolutely right.

Raycoms (dismissed):

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -106,13 +109,18 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
...
-                hut = "";
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
+                {
+                    section = hut;
+                    break;
+                }
+                else
+                {
+                    hut = "";
+                }
```
marchermans: If this is meant to work for other mods, then we should change this mechanism.
Use some form of marker interface instead of using a `blockhut` prefix. or a map that is named `HUT` since this is then really not clear.

marchermans (changes requested):

src/main/java/com/ldtteam/structurize/management/StructureName.java
Raycoms: @ShadowProtocol is doing a bigger rework with propper metadata files which will contain the propper descriptions for the future. But that rework will take a while longer. This is meant to work for the time being as a replacement for @Justen-N addon to work propperly.

marchermans: That has nothing todo with a rework.
This just needs to be defined better.

So no. Make it a marker interface on the block and call it a day.

Justen-N: This was a 'quick fix' (read: bad idea) for a straight-across companion mod ( herder compatibility) But I can see where you are coming from.

Justen-N: I guess, my question regarding a marker interface, is where would we put it? In Structurize? It feels like that would force an even tighter coupling between Minecolonies and Structurize.

Justen-N: I suppose we could check if the block extends the AbstractBlockHut class, if that works for you.

Justen-N: By the way, at the moment, all of minecolonies hut blocks still require the blockhut prefix, which is why I did not change that portion of the check. If we remove it, we'll have to refactor the blocks from Minecolonies.

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -106,13 +108,18 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
...
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
-            {
-                section = hut;
-            }
-            else
+            //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
+            for (String origin : StructureLoadingUtils.getOriginMods().keySet())
             {
-                hut = "";
+                if (IAnchorBlock.class.isAssignableFrom(ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)).getClass()))
```
Nightenom: IAnchorBlock is definitely not a way for me to distinguish huts and the others
@OrionDevelopment meh, reworking this to something "normal" does not make much sense - the mod is one gigantic tape fix and every time you try to rewrite something you end up rewriting half of the mod, I'm completely fine with previous minor change

Justen-N: I'm willing to revert this. My reasoning was that every hut block implements IAnchorBlock. I do realize that decoration controllers do so as well, but figured that they would be placed elsewhere by StructureLoadingUtils because of the path used .

Justen-N:
@OrionDevelopment, I'm leaving this in the state you originally reviewed it in. Changing this to be a 'true' implementation is a loss of effort that could be put in elsewhere, especially because this check will need to be changed completely alongside the metadata changes that are planned. For the moment, the simple solution proposed here is, in effect, correcting for technical debt in the short term, until that debt can be fully addressed later on.

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -106,13 +109,18 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
...
-            }
-            else
-            {
-                hut = "";
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
+                {
+                    section = hut;
+                    break;
+                }
+                else
```
Nightenom: This is not a correct logic rewrite btw, if you iterate first then you end up with hut = ""

Justen-N: ... Did I seriously.... damn.

Justen-N: Alright, I just removed the else block. From what I can see, it isn't necessary anyway.

Nightenom: Even now it's not correct :D you are missing the else branch logic now

Nightenom: And also the list is probably needed? Depends on usage in mcol

Justen-N: .... Wow. I'm just gonna go sit in a corner now.

Justen-N: @Nightenom, the HUTS list is not referenced from minecolonies. I'm pretty sure it was originally populated so that it could be checked against by the BuildTool. I went through each time StructureName is imported by mcol, it's never used.

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -106,13 +108,18 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
         {
             hut = schematic.split("\\d+")[0];
             section = Structures.SCHEMATICS_PREFIX;
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
+            //This allows mods other than minecolonies to create hut buildings and use the buildtool with them.
+            for (String origin : StructureLoadingUtils.getOriginMods().keySet())
             {
-                section = hut;
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + hut)) != Blocks.AIR)
```
marchermans, line 114: Again this won't be merged in the current state.
I talked this through with Ray.

Since Structurize 2.0 is a long long long long long way away, we have to work with this.
And since a standing (partially unwritten) rule for this mod is that as long as it is for Minecolonies then everything is allowed. But since PR in particular is targeted at supporting other mods, this won't fly and the exception does not take here.

In other words, create a proper API for this instead of hacking around.

marchermans (changes requested):

src/main/java/com/ldtteam/structurize/management/StructureName.java
Raycoms: I wrote in the other side. Let's remove all the hut stuff from this StructureName class, make in MineColonies an implementation that overrides this Structurename and handle it there.

marchermans: Also possible.

Justen-N:
I talked with ray as well, and agreed that we will handle this on the
minecolonies side. I'm out of town for a week, but will pick this thread up
later on.

PR #411, opened 2021-07-20
StructureName API interface

Justen-N:
I'm back to annoy you again. Decided that while putting what cares about a hut is or isn't is a Minecolonies side thing, Orion was right that if we want a different mod to be able to define a section, what all of my changes revolve around, adding a marker interface and implementing it to MC and other mods was the better choice.

- Refactor references to 'hut'(Minecolonies specific) to the more general 'sectionID'
- Create ISectionDefinition Interface, to allow mods to custom define the section a block belongs to. 
Questions: 
- Currently, all minecolonies blocks are defined with 'blockhut' and their respective name/section. Should this be changed now or wait for a future version? Changing it now makes implementing sections/buildings for other mods with structurize easier, but could cause a breaking change in Minecolonies, as all the hutblocks would need to be refactored. 
- JAVADOC Review comments concerning the isHut and getHutName functions
Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/management/StructureName.java
+++ b/src/main/java/com/ldtteam/structurize/management/StructureName.java
@@ -1,10 +1,13 @@
 package com.ldtteam.structurize.management;
 
+import com.ldtteam.structurize.api.blocks.ISectionDefinition;
 import com.ldtteam.structurize.util.LanguageHandler;
 import com.ldtteam.structurize.api.util.constant.Constants;
+import com.ldtteam.structurize.util.StructureLoadingUtils;
 import net.minecraft.block.Blocks;
 import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.RegistryManager;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.ArrayList;
@@ -18,7 +21,6 @@ import java.util.regex.Pattern;
  */
 public class StructureName
 {
-    public static List<String> HUTS = new ArrayList<>();
 
     private static final Pattern levelPattern              = Pattern.compile("[^0-9]+([0-9]+)$");
     private static final String  LOCALIZED_SCHEMATIC_LEVEL = "com.ldtteam.structurize.gui.buildtool.hut.level";
@@ -29,13 +31,14 @@ public class StructureName
     private              String  prefix                    = "";
     private              String  style                     = "";
     private              String  schematic                 = "";
-    private              String  hut                       = "";
+    private              String  sectionID                       = "";
 
     /**
      * Create a StructureName object from a schematic name.
      *
      * @param structureName as huts/stone/Builder1 or decorations/Walls/Gate
      */
+
     public StructureName(@NotNull final String structureName)
     {
         init(structureName);
@@ -102,17 +105,16 @@ public class StructureName
 
         //The section is the prefix, except fot hut
         section = prefix;
-        if (prefix.equals(Structures.SCHEMATICS_PREFIX))
-        {
-            hut = schematic.split("\\d+")[0];
+        if (prefix.equals(Structures.SCHEMATICS_PREFIX)) {
+            sectionID = schematic.split("\\d+")[0];
             section = Structures.SCHEMATICS_PREFIX;
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
-            {
-                section = hut;
-            }
-            else
-            {
-                hut = "";
+            for (String origin : StructureLoadingUtils.getOriginMods().keySet())
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + sectionID)) instanceof ISectionDefinition) {
+                    section = sectionID;
+                }
+            if (section != sectionID) {
+                sectionID = "";
+
             }
         }
     }
@@ -122,10 +124,11 @@ public class StructureName
      * such as Builder, Citizen, ...
      *
      * @return the name of the hut.
+     * REVIEW: deprecate or change this method?
      */
     public String getHutName()
     {
-        return hut;
+        return sectionID;
     }
 
     /**
@@ -193,15 +196,16 @@ public class StructureName
     }
 
     /**
-     * Whether the schematic is a huit or not.
+     * Whether the schematic is a hut or not.
      * This is done using the naming convention only, should start by huts/
      * and a structurize block name should exist.
      *
-     * @return True is it is a hut otherwise False
+     * @return True if it is a hut otherwise False
+     * REVIEW: deprecate or change this method?
      */
     public boolean isHut()
     {
-        return !hut.isEmpty();
+        return !sectionID.isEmpty();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Justen-N:
@OrionDevelopment, I assume this is closer to what you meant.

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -102,17 +105,18 @@ else if (name.startsWith(Structures.SCHEMATICS_CACHE + Structures.SCHEMATICS_SEP
...
         section = prefix;
-        if (prefix.equals(Structures.SCHEMATICS_PREFIX))
-        {
-            hut = schematic.split("\\d+")[0];
+        if (prefix.equals(Structures.SCHEMATICS_PREFIX)) {
+            sectionID = schematic.split("\\d+")[0];
             section = Structures.SCHEMATICS_PREFIX;
-            if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(Constants.MINECOLONIES_MOD_ID, "blockhut" + hut)) != Blocks.AIR || HUTS.contains(hut))
+            for (String origin : StructureLoadingUtils.getOriginMods().keySet())
+                if (ForgeRegistries.BLOCKS.getValue(new ResourceLocation(origin, "blockhut" + sectionID)) instanceof ISectionDefinition)
```
Raycoms, line 112: I'm not 100% happy about the "blockhut" part of this solution though.

Justen-N: Neither am I. I can push it all the way through to remove the blockhut convention, but that means a breaking change for MC, as we would have to change all of the blockhut registered blocks to the new standard, whatever we decided that was. I would probably just go with the building name, but that's my preference.

uecasm: Maybe `HUTS` could be made into a `ResourceLocation` list (or some kind of registry), and then it could be left to MineColonies and other mods to register their hut blocks with Structurize (via an API method, not by directly touching `HUTS`) when they want them to be treated as huts?  Currently it looks like `HUTS` isn't used by anything anyway.

Or probably MineColonies could do it by itself -- it can walk the building registry and get all the hut blocks from all mods, so the other individual mods only need to register a building with MineColonies and don't need to register the hut with Structurize.

ShadowProtocol:
There was a discussion I think I missed somewhere, so I don't have the full picture here, but just an FYI ... 
I'm in the process of moving style and some structure data to datapacks, which might change how this works anyway, I haven't gotten to this particular integration yet so I don't know the depth at which I'll need to adjust things.
I'm aiming to have all of my changes done by the start of August.

Raycoms:
We have to think about a good way to categorize schematics, and a good way how to detect huts. We could go the "include the anchorblockname into the name of the schematic" way, which would then be guardtower1_blockhutguardtower for example, and then we can query the block and check sth on the interface. Or have some propper metadata in general that allows us to detect things without opening the file. As

xml with mapping and we read the xml on the minecolonies side instead of detecting it through the structurename

Justen-N:
Shadow, that was why I wanted to leave my 'little, hacky' fix in so that I could get something moving, and then your changes would be one step closer to something more compatible, but Orion said do it right, which is why we're here. I'm willing to work out the metadata, but that requires waiting for the datapack conversion. I'm probably just going to stop trying to get something in the main project for my addon for now. I'll just run my 'hack' on my fork and keep building over there until I can contribute here.

uecasm:
> Or have some propper metadata in general that allows us to detect things without opening the file.

You _could_ read some metadata out of the file -- at load time you're calculating a file hash anyway, so you're already opening every file and reading them into memory.  Decompressing and NBT parsing isn't going to add that much overhead on top of that.

But like I said earlier, it'd be easiest to just make mcol iterate the building registry and register all known hut blocks and/or hut schematic names up front, especially since that will take care of all addon buildings too.  (The only missing link is that the BuildingEntry would need to know the schematic name directly instead of delegating that to the Building, but that's probably better anyway.)

Raycoms:
> > Or have some propper metadata in general that allows us to detect things without opening the file.
> 
> You _could_ read some metadata out of the file -- at load time you're calculating a file hash anyway, so you're already opening every file and reading them into memory. Decompressing and NBT parsing isn't going to add that much overhead on top of that.
> 
> But like I said earlier, it'd be easiest to just make mcol iterate the building registry and register all known hut blocks and/or hut schematic names up front, especially since that will take care of all addon buildings too. (The only missing link is that the BuildingEntry would need to know the schematic name directly instead of delegating that to the Building, but that's probably better anyway.)

Oh, I didn't know that the hash calculation required to open each file already. If that is the case we can add header metadata to each file.

Justen-N:
> There was a discussion I think I missed somewhere, so I don't have the full picture here, but just an FYI ...
> I'm in the process of moving style and some structure data to datapacks, which might change how this works anyway, I haven't gotten to this particular integration yet so I don't know the depth at which I'll need to adjust things.
> I'm aiming to have all of my changes done by the start of August.

Not to be a nudge, but how is this going @ShadowProtocol ?

ShadowProtocol:
So it turns out I took the complicated route, and I'm dredging up a lot of the code that searches through schematics. the ``StructureName`` class also had its tendrils through that system, and in the end the altered system will likely not use that class at all. I'll need to work with things a bit more to see how it all will come together.

But if what you have here is satisfactory for a temporary solution I agree we should just go with this for now. My plans may change again, and I have a feeling there'll be a bit of back and forth in the review for it.

Raycoms:
> So it turns out I took the complicated route, and I'm dredging up a lot of the code that searches through schematics. the `StructureName` class also had its tendrils through that system, and in the end the altered system will likely not use that class at all. I'll need to work with things a bit more to see how it all will come together.
> 
> But if what you have here is satisfactory for a temporary solution I agree we should just go with this for now. My plans may change again, and I have a feeling there'll be a bit of back and forth in the review for it.

If we could get rid of the StructureName class that'd definitely be a win.

Justen-N:
@OrionDevelopment, @Raycoms: Do we implement this fix? Especially since we can sneak in the BlockHut convention removal during the 1.17 update. Again, it isn't the big fix we're all hoping for, but it will work for now.

TagnumElite:
I have been thinking about this and how to go about it.
I wrote most of my [thoughts is a gist here](https://gist.github.com/TagnumElite/eb4c3fadc5d95108da815533430d5cd7).

The current method structurize uses is a bit too limited right now and I dislike using mixins when I don't have too.

Nightenom:
> I have been thinking about this and how to go about it. I wrote most of my [thoughts is a gist here](https://gist.github.com/TagnumElite/eb4c3fadc5d95108da815533430d5cd7).
> 
> The current method structurize uses is a bit too limited right now and I dislike using mixins when I don't have too.

Commenting first sentence, any packs were problem in past. Now datapacks might seem useable but still I (personally) would like to avoid using in jar blueprints for anything but extracting them on first load to normal folders (at OS level).

The rest is nice but most likely not extendable design that fits mostly mcol use only.

TagnumElite:
1. I not saying that modpack problems still existed. I was just saying that using datapacks, you could disable schematic definitions using `{}` in the file.
2. Schematics can still be extracted, just either extract the files on request or on load.
3. You specify that using datapacks is not extensible, is this because that this method would require custom datapacks by the user instead of them just add a blueprint file to the `structurize` folder? (If so, then we can still keep the structurize folder with some modified functionality. If not, then please explain so that I may better understand the goal.)

Raycoms:
We were fine with this one: https://github.com/ldtteam/Structurize/pull/452 ? @Nightenom?

Raycoms:
Already solved this otherwise

PR #412, opened 2021-07-21
Several fixes

Raycoms:
- Fixes an issue with optifine line rendering
- Fixes an issue with the inwardcircle iterator
- Adds a random iterator.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/BlueprintIteratorRandom.java
+++ b/src/main/java/com/ldtteam/structurize/placement/BlueprintIteratorRandom.java
@@ -1,0 +1,101 @@
+package com.ldtteam.structurize.placement;
+
+import com.ldtteam.structurize.placement.structure.IStructureHandler;
+import net.minecraft.util.math.BlockPos;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+
+/**
+ * This is the default blueprint iterator.
+ * It's a helper class used to track the progress of one iteration over the structure.
+ */
+public class BlueprintIteratorRandom extends AbstractBlueprintIterator
+{
+    /**
+     * Random ordered list of positions.
+     */
+    final List<BlockPos> positions = new ArrayList<>();
+
+    /**
+     * Initialize the blueprint iterator with the structure handler.
+     * @param structureHandler the structure handler.
+     */
+    public BlueprintIteratorRandom(final IStructureHandler structureHandler)
+    {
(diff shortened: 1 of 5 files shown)


src/main/java/com/ldtteam/structurize/placement/BlueprintIteratorRandom.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+
+/**
+ * This is the default blueprint iterator.
+ * It's a helper class used to track the progress of one iteration over the structure.
+ */
```
uecasm, line 14: A better javadoc would be good.

uecasm (dismissed):
I didn't verify the math/logic in detail, but seems generally reasonable otherwise.  Although I doubt players would want to use a random iterator, though it might be useful for annoying the pathfinding code as much as possible (and perhaps discovering problems, albeit hard-to-reproduce ones).

Raycoms:
> I didn't verify the math/logic in detail, but seems generally reasonable otherwise. Although I doubt players would want to use a random iterator, though it might be useful for annoying the pathfinding code as much as possible (and perhaps discovering problems, albeit hard-to-reproduce ones).

It was actually quite interesting =D At least for smaller buildings =D

uecasm (dismissed):

src/main/java/com/ldtteam/structurize/util/RenderUtils.java
```diff
@@ -175,7 +175,7 @@ public RenderTypes(final String nameIn,
             RenderType.State.builder()
                 .setLineState(new RenderState.LineState(OptionalDouble.empty()))
                 .setLayeringState(VIEW_OFFSET_Z_LAYERING)
-                .setTransparencyState(GLINT_TRANSPARENCY)
+                .setTransparencyState(NO_TRANSPARENCY)
```
Nightenom, line 178: `optifine compat ? no_trans : glint_trans`
I remember this being problem with some shaders, I also remember creating a fix for that - not sure where it went tho :D
Root cause is probably this being rendered in wrong framebuffer (which you really don't need to care about, will fix it once forge renderWorldEvents get redesigned)

Nightenom (dismissed):

src/main/java/com/ldtteam/structurize/util/RenderUtils.java
Raycoms: but do we need the glint transparency in the first place here?

Nightenom: I chose this because entity outlining is also glint, so the player gets that it's just outline and not something special

someaddons (approved):

PR #415, opened 2021-08-03
Block Optimizations

WenXin20:
- Optimizes timber frames by adding cullfacing to the parent models. (Main change)
- Fixes shingles with z-fighting and a slight misalignment with the shingles slab.
- Adds a few texture improvements, but those can be removed if they aren't preferred.
- Renames barrels. (From "Barrel Horizontal" to "Horizontal Barrel")

New: 
 

Old: 

New:

Old:

Review please


Raycoms:
1.16 is eol

PR #429, opened 2021-09-03
Adds two general utility schematics

uecasm:
- Adds `fill`, an 8x8x8 of solid placeholders (to fill an area with solid blocks).
- Adds `dig`, an 8x8x8 of air (to dig out or clear an area).

Review please

Inspired by a conversation with Shindria on Discord.


Nightenom:
Idc, but I would rather see generic fill and clear buttons in scan tool. Also can't you just use shape tool for this already?

uecasm:
Shape tool (and scan tool) is creative only; this is useful for MineColonies survival as well.  (I was going to put it there at first, but it seemed worthwhile making it more general.)

Nightenom:
Then it might be better to add them into mcol

Raycoms:
The general idea was to add shape-tool support to MineColonies. So that in 'non creative" a build order for the builder is done.

uecasm:
That could indeed be more useful, but until that happens these seem like a good interim method.  I can put them in an mcol PR instead if you prefer.  Or would you rather drop the idea entirely?

Raycoms:
> That could indeed be more useful, but until that happens these seem like a good interim method. I can put them in an mcol PR instead if you prefer. Or would you rather drop the idea entirely?

A MineColonies PR would be okay. Integrating it should not be very hard either as the shapetool already handles it with blueprints too.

PR #432, opened 2021-09-03
Porting cleanup

Nightenom:
- removed a lot of unreferenced methods and constants
- transformed few classes to records
- guis are working, rendering needs more proof testing tomorrow

Review please


src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
```diff
@@ -172,25 +159,7 @@ public static void onClientTickEvent(final ClientTickEvent event)
...
-            Minecraft.getInstance().getProfiler().pop();
-        }
 
         Minecraft.getInstance().getProfiler().pop();
     }
-
-    /**
-     * Used to catch the scroll when no gui is open.
-     *
-     * @param event the catched event.
```
Raycoms, line 188: Don't we need this anymore?

Raycoms:
Looks fine overally. Did you test this with a few different things? (Different schematics, shape tool, scan tool replacement UI, etc)

src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
Nightenom: its in blockui

Raycoms: ah okay

Raycoms:
Tested and works. I'm not 100% with the border rendering yet. But it's good enough to merge. I believe that might even be on blockUI side.

Raycoms:
Just noticed entity rendering is still broken.

Raycoms:
I will merge this as it fixes a series of other things. I did some debugging and I got closer to why entities render so messed up. 
Doing this: 

```
        RenderSystem.getModelViewStack().pushPose();
        RenderSystem.getModelViewStack().mulPoseMatrix(matrixStack.last().pose());
        RenderSystem.applyModelViewMatrix();
```

before the entity rendering will double apply the viewpoint translations to the entities. Weirdly enough, a bunch of other things still seriously misbehave.

Nightenom:
Well, you know, draft prs are drafts for reason, and also this:
>  rendering needs more proof testing tomorrow

Nightenom:
> Tested and works. I'm not 100% with the border rendering yet. But it's good enough to merge. I believe that might even be on blockUI side.

Well, mojang rendering sucked, what you see is that the box border is now properly scaled while the old one was immune to gui scaling

Raycoms:
So we just have to make the lines smaller by default now

Sent from ProtonMail mobile

PR #435, opened 2021-09-04
Feature/shape tool integration

uecasm:
- Adds rotation/mirroring indicators recently added to build tool also to shape tool
- Makes some preparatory changes to allow the shape tool to be integrated with MineColonies
    - One side effect of this is a paste button similar to the build tool, but in pure Structurize this behaves identically to the tick.
    - Another side effect is a new chat message shown if you try to use the tool in survival mode.
    - This also fixes a nasty bug in the way that the shape tool (or anything other than the scan tool) generated blueprints.
- Adds +/- buttons to the shape dimensions for ease of use.
- Fixes the </> buttons to the shape type that already existed but did nothing.
- Removes some "todo remove if not used" code that was not used.
- Hides the preview after pasting a shape (rather than leaving it), but allows the same shape to be quickly recalled in another location.

Review please

image


Raycoms (approved):

src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
```diff
@@ -230,7 +231,7 @@ public void addBlockState(final BlockPos pos, final BlockState state)
 
         if (index == -1)
         {
-            index = this.palleteSize + 1;
+            index = this.palleteSize;
```
Nightenom, line 234: how is this bug when it would affect the blueprints immediately? I don't see any way  how this would not break the game already since it affects the raw saved data

src/main/java/com/ldtteam/structures/helpers/Settings.java
```diff
@@ -309,7 +309,7 @@ public void setActiveSchematic(final Blueprint blueprint)
     {
         if (blueprint == null)
         {
-            reset();
+            resetBlueprint();
```
Nightenom, line 312: reason?

src/main/java/com/ldtteam/structurize/client/gui/WindowShapeTool.java
```diff
@@ -541,21 +668,27 @@ private void rotateLeftClicked()
...
+
     /**
      * Cancel the current structure.
      */
     private void cancelClicked()
     {
-        Settings.instance.reset();
-        Network.getNetwork().sendToServer(new LSStructureDisplayerMessage(null, false));
-        close();
+        clearAndClose();
```
Nightenom, line 691: no need for alias method

src/main/java/com/ldtteam/structures/blueprints/v1/Blueprint.java
uecasm: Shape tool blueprints weren't saved (before).  Scan tool blueprints use the other constructor, which calculated it correctly.

The shape tool preview didn't crash because there's the reverse of this bug down in `addBlockState` that caused it to calculate the correct index (based on the wrong initial state).  However the `palletteSize` was always too small, so if you actually tried to save that blueprint it would miss one block from the palette and would crash whoever tried to load it.

TBH the whole `palleteSize` field seems entirely unnecessary (and only a potential source of bugs).  It'd be better to get rid of it entirely, but I didn't want to make too many changes at once.

src/main/java/com/ldtteam/structurize/client/gui/WindowShapeTool.java
uecasm: There is, because it's called from more than one place.  True, the cancel handler *could* call that directly but that is poor style.

src/main/java/com/ldtteam/structures/helpers/Settings.java
uecasm: The `blueprint` field needs to be cleared otherwise the preview remains visible after closing the GUI.

However, I didn't want to clear the other fields, so that you can quickly place multiple copies of the same shape.  (It does still lose the mirror/rotation somewhere or other, but all other properties are preserved.)

src/main/java/com/ldtteam/structurize/client/gui/WindowShapeTool.java
Nightenom: Like you literally extracted content of this method to new method which does exactly the same

src/main/java/com/ldtteam/structures/helpers/Settings.java
Nightenom: Yeah, but this method is called from different places and you weakened its effect without making it up elsewhere

src/main/java/com/ldtteam/structurize/client/gui/WindowShapeTool.java
uecasm: It doesn't, actually.  I wrote a new method that does something slightly different, and then called it from three places.  One of those places happens to be the cancel button click handler, which also happens to not do anything else, but that isn't a good reason to call a utility method directly as the cancel button click handler instead of actually having a named cancel button click handler.  It's possible that in the future some other action may be needed on cancel that doesn't happen for the other cases; but regardless of that, method names matter (and the other two places should definitely not be calling a "cancel" method).

In fact arguably cancel could be a good time to do a full reset (back to single gold cube), in which case it would be different from the other two cases -- but on the other hand it seems more user-friendly to still retain the shape in that case, to not punish people Xing out just to move to a different position.

src/main/java/com/ldtteam/structures/helpers/Settings.java
uecasm: It's called from exactly one other place, where it is considered an error for blueprint to be null (and is very unlikely, though not impossible).  It seemed relatively harmless.  And it seems incorrect to clear all the other properties from a blueprint-setter method.

src/main/java/com/ldtteam/structurize/client/gui/WindowShapeTool.java
Nightenom: Having an alias has no good reason either, also you named method updateRotationState and call it from mirror method - how does that make sense I don't know, and future code can arise in day as well as ten years. simply having alias method has no valid reason in my eyes, clearAndClose can surely be target method for button handler

src/main/java/com/ldtteam/structures/helpers/Settings.java
Nightenom: Well I would rather see a softreset method for shapetool but ok

src/main/java/com/ldtteam/structurize/client/gui/WindowShapeTool.java
uecasm: No, it can't.  That's just silly.  I've already explained why.

Code readability is vastly more important than dubiously saving one method call, especially in a non-performance-critical code path that happens once in direct response to user input.

PR #436, opened 2021-09-11
Feature/groundlevel

uecasm:
A better way to implement ldtteam/minecolonies#7588

- Adds a "Tag Anchor" block, which acts as an anchor block and can be used to hold Tag Tool tags when MineColonies is not loaded, or when making a decoration without a MineColonies Decoration Controller.  On fancy placement it will always convert to air (so it can only be used for temporary tags, but is safe to rescan provided you only use the left paste).
- Modifies the `big_well` default schematic to include a tag anchor and `groundlevel` tag.
    - Also removes the MineColonies waypoint that it previously erroneously had, and was auto-upgraded to the new shingles, plus a few other side-effects of rescanning.
    - Note that currently the tag anchor is slightly off-center, and so the well will rotate around that rather than around the blueprint center.  I don't consider this a problem, but it could easily be fixed if you think it is.
- Modifies the Build Tool UI so that it automatically offsets from the current anchor position to the ground level of the active schematic (such that as you switch between schematics it will always correctly align the ground level -- assuming that the player originally clicked on a block in the ground, as is usually the case when placing buildings).
    - If the player explicitly moves the preview up or down, then it will maintain that same relative placement as you switch between schematics.

Review please

Side note: I lifted the implementation of `BlueprintTagUtils` from the other PR, but while testing this it didn't work as written, it was reading the wrong NBT compound.  It's possible this was something I introduced rather than a bug in the original PR, but I didn't see any obvious mistakes in either my code or the PR refactoring; unless this just never worked, or I missed something.


Nightenom (dismissed):

src/main/java/com/ldtteam/structurize/client/gui/WindowBuildTool.java
```diff
@@ -881,6 +886,28 @@ private void changeSchematic()
         {
             Settings.instance.setPosition(this.pos);
         }
+
+        adjustToGroundOffset();
+    }
+
+    private void adjustToGroundOffset()
```
someaddons, line 891: javadoc

src/main/java/com/ldtteam/structurize/client/gui/WindowBuildTool.java
```diff
@@ -881,6 +886,28 @@ private void changeSchematic()
         {
             Settings.instance.setPosition(this.pos);
         }
+
+        adjustToGroundOffset();
```
someaddons, line 885: I think for usability we should not adjust the height when only the schematic level changes, would be annoying having to offset the Y again when switching through the levels

uecasm: If the ground level doesn't change between the levels then there will be no change, it subtracts the previous offset and then adds the new offset, resulting in a net zero difference in the position if they were the same.  This will work regardless of whether they're switching between levels or switching between different schematics that happen to be the same.

src/main/java/com/ldtteam/structures/client/BlueprintRenderer.java
```diff
@@ -112,7 +112,8 @@ private void init()
                 try
                 {
                     BlockState state = blockInfo.getState();
-                    if (state.getBlock() == ModBlocks.blockSubstitution.get())
+                    if (state.getBlock() == ModBlocks.blockSubstitution.get() ||
+                        state.getBlock() == ModBlocks.blockTagSubstitution.get())
```
Raycoms, line 116: So we want to render the solid placeholders as air?

Raycoms (dismissed):
Do tags work with rotate too?

uecasm:
> Do tags work with rotate too?

I didn't change any of that, but yes, they do.

someaddons (approved):

PR #439, opened 2021-09-16
Another porting progress

Nightenom:
- Let BE model data disappear after block baking, no need to hold it during rendering
- Breaking change for mcol: rework line rendering, relocate and replace few buffers
- Fix entity rotating and mirroring and placing, things like armor stand and paintings do now work properly, the only thing which does not work is mirroring of paintings with `width > 1`, hopefully tested enough
- Leash is pita to fix, unless needed leashes won't work

Review please


Raycoms (dismissed):

Nightenom:
Mergeable but doubt you can fix mcol site, will do myself probably during the weekend

Raycoms (approved):

marchermans (approved):
This indeed looks good.
Shipit!

PR #442, opened 2021-09-20
Ignore non-serializable entities (such as players) during scan

uecasm:
Related to ldtteam/minecolonies#7605

- Don't bother saving non-serializable entities to save some disk space (they won't get placed anyway)

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java
+++ b/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java
@@ -122,6 +122,7 @@ public class BlueprintUtil
         {
             final Vector3d oldPos = entity.position();
             final CompoundNBT entityTag = entity.serializeNBT();
+            if (!entityTag.contains("id")) continue;
 
             final ListNBT posList = new ListNBT();
             posList.add(DoubleNBT.valueOf(oldPos.x - pos.getX()));


Raycoms (approved):

marchermans (approved):

Nightenom:
Might be better to use the 1.17 fix if applicable in 1.16, else to Ray, do not port this commit
 <https://github.com/ldtteam/Structurize/commit/62079fc82b70e4276b6c482727465570f8723765#diff-f7374cafd08e0b43f6580ddc56163832850f7d5bf80b2aad08b7c3e81c3c2a17R122>

uecasm:
> Might be better to use the 1.17 fix if applicable in 1.16

I did consider checking `entity.getType().canSerialize()` (it is valid in 1.16 too) but this method is better.  There are more reasons than just that which can cause the entity to not serialize an `id`, which is what really matters.  So you should replace that 1.17 code with this change, not the reverse.

Or possibly do both things -- checking `canSerialize` up front will save a bit of processing time.  Probably not enough that anyone will care, though.

Nightenom:
In 1.17, there is no way id would not serialise if `canSerialize` returns true. And the method is final. And how would you deserialise when you are not saving id, so if someone manages not to register their type correctly, it's their fault and I prefer not to hide others bugs. Not sure what the method looks like in 1.16 though, and I do agree that contains id is more safe  

image

PR #443, opened 2021-09-28
Fix groundlevel again

uecasm:
- Turns out Structurize and MineColonies used different (and inconsistent) methods to choose the initial block to anchor the build (and shape) tool previews at.  (Partly because they have to hook different events to do so.)  This changes Structurize to use the same choice as MineColonies.
- Specifically, the initial anchor is the block adjacent to the clicked face of the clicked block.  For the shape tool, this lets you build against a surface nicely.  For the build tool, the assumption is that most of the time you will be clicking the top face of the "ground" block anyway.  (And when the hut/anchor block is sitting directly on top of the ground, then this is already the correct location.)
- This also adjusts the treatment of schematics without a groundlevel tag to assume that the ground level is one block below the anchor (which agrees with the above assumption, and with the wiki docs for the groundlevel tag).  This is not true of all existing schematics, but there is no rule that would work universally.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structures/lib/BlueprintTagUtils.java
+++ b/src/main/java/com/ldtteam/structures/lib/BlueprintTagUtils.java
@@ -4,10 +4,12 @@ import com.ldtteam.structures.blueprints.v1.Blueprint;
 import com.ldtteam.structurize.blocks.interfaces.IBlueprintDataProvider;
 import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.util.math.BlockPos;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.List;
 import java.util.Map;
 
+import static com.ldtteam.structurize.api.util.constant.Constants.GROUNDLEVEL_TAG;
 import static com.ldtteam.structurize.blocks.interfaces.IBlueprintDataProvider.TAG_BLUEPRINTDATA;
 
 /**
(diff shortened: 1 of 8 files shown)


uecasm:
Investigating some issues with supply camps and ships.

uecasm:
This is less straightforward than I was originally hoping, but I've tested it and it works with all MineColonies build sources.

The supply camp/ship use a different rule for ground alignment (only when built using their dedicated item -- to get them working properly using the build tool they will need to be rescanned with a groundlevel tag, since its rule is different).  More details in the mcol PR.

Raycoms (approved):

src/main/java/com/ldtteam/structures/lib/BlueprintTagUtils.java
```diff
@@ -43,4 +45,59 @@ public static BlockPos getFirstPosForTag(final Blueprint blueprint, final String
...
+     * @return The number of levels, e.g. 1 means that the bottom layer of the
+     *         blueprint is ground and 3 means the bottom 3 layers are ground.
+     */
+    public static int getNumberOfGroundLevels(@NotNull final Blueprint blueprint,
+                                              final int defaultGroundLevels)
+    {
+        final BlockPos groundLevelPos = getFirstPosForTag(blueprint, GROUNDLEVEL_TAG);
+        if (groundLevelPos != null)
+        {
+            return blueprint.getPrimaryBlockOffset().getY() + groundLevelPos.getY() + 1;
```
someaddons, line 63: This one seems weird to me: E.g. we got the anchor/prim offset at Y=5, the tag Y-level is at 7 and then the groundlevels are from 0 to 13?

uecasm: The tag y-level is relative to the anchor.  So `groundLevelPos.getY() == -1` where the ground is one level below the anchor, for example.

someaddons: ah right nvm :D

someaddons (approved):

PR #477, opened 2022-01-16
Bans homicide (and suicide)

uecasm:
- Prevents players in creative mode from being able to delete other players (or themselves) for the crime of wandering into an active scan area.
    - (Also fixes it for a couple of other entity types, to match the [scan-time filter](https://github.com/ldtteam/Structurize/blob/30fb6a9c9633caf479130c934071551f04038734/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java#L123).  Though those are far less likely to be inside the scan area.)

Review please

Note that 1.16 has a slightly different [scan-time filter](https://github.com/ldtteam/Structurize/blob/88c4efa78b119dc14978d08e094515903cd85dc8/src/main/java/com/ldtteam/structures/blueprints/v1/BlueprintUtil.java#L125), but the effect is similar, so this may want to be backported.  (And perhaps the filter changed to match, although they're equivalent.)

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/gui/WindowScan.java
+++ b/src/main/java/com/ldtteam/structurize/client/gui/WindowScan.java
@@ -25,7 +25,6 @@ import net.minecraft.util.Tuple;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.decoration.GlowItemFrame;
 import net.minecraft.world.entity.decoration.ItemFrame;
-import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.entity.vehicle.AbstractMinecart;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
@@ -345,7 +344,7 @@ public class WindowScan extends AbstractWindowSkeleton
                     for (final Entity entity : list)
                     {
                         if (!entities.containsKey(entity.getName().getString())
-                                && !(entity instanceof Player)
+                                && entity.getType().canSerialize()
                                 && (filter.isEmpty() || (entity.getName().getString().toLowerCase(Locale.US).contains(filter.toLowerCase(Locale.US))
                                     || (entity.toString().toLowerCase(Locale.US).contains(filter.toLowerCase(Locale.US))))))
                         {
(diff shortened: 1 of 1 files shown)


Raycoms (dismissed):

src/main/java/com/ldtteam/structurize/client/gui/WindowScan.java
```diff
@@ -344,6 +344,7 @@ private void updateResources()
                     for (final Entity entity : list)
                     {
                         if (!entities.containsKey(entity.getName().getString())
+                                && entity.getType().canSerialize()
```
Nightenom: While I generally agree with this, can you exclude EntityType.LEASH_KNOT from this check? lighting bolt, player and fishing bobber are fine, but we want to list leash cuz currently we can't place them nor render them (in preview) properly - so builders should have way to check & remove them easily

uecasm: All the no-can-serialize entities are simply not saved at actual scan time, which means that they're effectively removed at that point anyway.  If there's a knot-to-knot leash then both ends go away; if there's a leash-to-animal then the animal might remain and be confused at the other end, but they'd appear in the list and that should be suspicious enough because animals shouldn't be in the scan either.

So I'm not sure why this isn't better as is -- it should just work without having to remember to do anything.  Can you explain further what the issue is?

Nightenom: Well, leash is normal entity that just has bullshit serialization, it's not unserializable entity in same way like bolt/bobber - it's serialized on the owning entity, thus if you place the entity with knot, that has wrong position information, then the knot just drops as item

and animals can be in any custom made scan - so that argument regarding them is just wrong

uecasm: Ok, but if someone ties an animal to a post and scans it, then builds that scan and the animal appears and a lead drops from them -- that's less ideal than if the lead were tied properly, but it still seems sufficiently correct; everything that was in the scan ended up back in the world, including the lead.  And if the player ignores it for five minutes then it ends up in the state you seem to be suggesting, where everything else is the same but the lead is gone.

Still, given that it does drop a lead item, you're right that it's probably worth explicitly alerting the player that there is a knot in the scan area, even if they don't remove it.  Having said that, it probably wouldn't be hard to make them actually work.  But that'd be a separate PR.

Nightenom: Making them work across whole structurize is bit pain, was looking into it few months ago. Need to look at how does vanilla templates solve that, if they don't file bug report to mojang and wait for them to make fix first (since it's not something game-breaking)

Nightenom (approved):

PR #480, opened 2022-01-22
Fix chests (and related) contents scan

uecasm:
Closes discord report

- Makes chests work ~properly~ again in 1.18
    - Added warning that container contents may not be listed
- When clearing the scan tool filter after entering some text, it wasn't switching back to show all items

Review please

This is specific to 1.18 (and perhaps 1.17, but I guess we don't care about that any more); this works properly on 1.16 already.

There were two underlying problems that this fixes:
1. ~Incorrect resource list on client side:  Chests don't appear to sync their items from server to client any more (at least not unless you have their inventory gui open).  This caused the scan tool resource list to omit any contents of chests or other containers, as this was performed on client side.  This has been fixed by making the scan window request the resource list from the server as needed.  (Caveat: due to the input event structure this can be quite noisy, and may need to be further optimised.)~
2. Incorrect resource list on server side;  The `ContainerHelper` class in 1.18 now restores the exact item layout, which means that you need to provide a list with the correct number of slots up front (but we were passing an empty list).  Since we don't want to assume a particular max size (maybe it's a super upgraded rack) this needs to interpret the contents manually.  This was affecting the Minecolonies Builder.


Nightenom (commented):
Generically speaking we at least want config switch (at server side) to forbid scanning (for the gui) on server threads. I'm partly aware of chest content not being in sync, but delegating entire scanning to server is a bit harsh.
Other changes are ok.

uecasm:
If you can think of a better way to get an accurate resource list, I'm all ears.  It _could_ be cached a little better than current implementation (so it doesn't requery on filter change), but changing either of the endpoints or closing and reopening the GUI would still probably need to force a requery.

And I suspect there might be some other BEs that only keep inventory or other data on server-side, so it's not necessarily just chests.

Nightenom (dismissed):

uecasm:
Per review, removed the client-server comms; the client-side resource list in the scan tool will just be incorrect for now.  Added a warning message that appears (regardless of scan contents) to indicate that container contents may not be listed.

Nightenom (approved):

PR #486, opened 2022-02-11
Wall mode

uecasm:
- ~gradle now caches even local builds to produce stale results when used as a dependency unless they're named -SNAPSHOT, so changed the default local build name accordingly~
- some consistency improvements when reloading blueprints after switching servers (or editing config and relogging)
- server-configurable custom build tool sections, e.g. ["halls;Hallways", "decorations/statues;Monuments to Okami"], where the left part is the directory prefix and the right part is the section name displayed in the left selector of the build tool (and remaining directory components, if any, are the style).
- mod-configurable custom build tool sections; ~currently just for Walls (though structurize itself does not include any wall blueprints by default)~
- internal refactoring to pass around PlacementSettings rather than individual rotation/mirror(/wall) settings
- ~Wall~ Extendable Mode: allows more easily creating long walls from smaller scanned segments
  - new UI appears in build tool ~when you have a /walls/ schematic selected (intended only for straight sections, though it can't really tell the difference at present)~
    - ~easiest selected by using the new Walls section heading, although not dependent on that~
    - appears when the schematic has an `extendable` tag on any block (typically the anchor itself, but doesn't really matter).  without Minecolonies, the Tag Anchor will have to be used for this.
  - automatically detects the orientation (axis) of the wall based on the longest horizontal size
      - can be overridden by applying the `extendaxis` tag on a block in the direction of the desired axis (mostly useful for square schematics)
  - wall can be "extended" (tiled) in either the positive or negative direction along the wall's axis
    - currently there is no particular limit on size, other than inherent memory limits; possibly there should be something for sanity reasons (and server trolls)
  - wall can be "overlapped" (spacing between tiles reduced) to allow for some wall schematics that don't quite tile cleanly as is
  - wall generation occurs independently on both server and client sides; when schematic sync is needed only the original un-tiled schematic is transferred
  - caveat: walls/stairs at the edges of the schematic will not always link up nicely (seems to work reasonably well when the builder places it, but not when pasted, though it's not entirely consistent). recommend that future wall styles have an out-of-schematic extra block so that the ones inside the scan have the right blockstate to join properly. otherwise maybe there needs to be an extra step to re-update fences/walls after a build?

Review please

1.18 only for now, though no particular reason it couldn't be backported later if we want.

image
(note this will no longer appear with Structurize alone, unless you put it in the config)

image
image


src/main/java/com/ldtteam/structurize/helpers/Settings.java
```diff
@@ -321,7 +325,7 @@ public void moveTo(final BlockPos pos)
     @Nullable
     public Blueprint getActiveStructure()
     {
-        return this.blueprint;
```
Raycoms, line 338: this seems like a weird change

Raycoms: why can't we re-use the blueprint?

uecasm:
Based on feedback, I've rebranded the UI for this to be "Extendable Mode" and made it use tags to detect when to show up, and also to override the axis auto-detection.  (Ray pointed out that this feature would be useful for roads/rails as well as walls.)

This also removes "Walls" as a default custom section, although keeps the custom-sections mechanisms in place (since mcol still wants them, and they could be useful for categorisation when used standalone still).  This removes some code that was mostly for mcol backwards-compatibility, although not completely pointless standalone.

I have not changed the code and NBT; that's still "wall mode" all the way down.  I can change that too if preferred, or perhaps a different name if a better one is decided.

gradle.properties
```diff
@@ -3,7 +3,7 @@ org.gradle.daemon=false
 
 modId=structurize
 modGroup=com.ldtteam
-modVersion=0.0.1-LOCAL
```
marchermans: No, for this purpose you can mark your dependency as changing, we don't use the `Snapshot` moniker since this has consequences for releases and other artifacts which are build locally.

src/main/java/com/ldtteam/structurize/helpers/WallExtents.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /** Deserialize from NBT */
+    public static WallExtents read(@NotNull final CompoundTag compound)
+    {
+        final short negative = compound.getShort("wallneg");
+        final short positive = compound.getShort("wallpos");
+        final short overlap = compound.getShort("wallovl");
+        return new WallExtents(negative, positive, overlap);
+    }
```
marchermans: This is not how Mojang uses NBT serialization.
Please follow the existing moniker and override the values in an existing instance.

marchermans (dismissed):
Generally okey, please refrain from changing CI and Build related options without consulting the relevant discord channels first, the `LOCAL` marker is on purpose.

gradle.properties
uecasm: That's not a good solution either, because Minecolonies doesn't do that.  It also doesn't work without some additional code that's also not in the build scripts.  At least not when I tried it.

Releases etc should be overriding that build version on the command line anyway, so it shouldn't matter.

And I did mention this change on Discord first.  Nobody objected.  It just works, with no fanfare.  This makes it a more sensible default.  If release builds etc need to do something more complicated to be correct, that's correct -- they're scripted, it's easier for them to be more complicated.

src/main/java/com/ldtteam/structurize/helpers/WallExtents.java
uecasm: I did do that at first, but it makes the call sites a lot more ugly for no good reason.  It's also inconsistent with the network deserialization.

uecasm: TBF, the instance method design does make more sense in the contexts where Mojang uses it, for block entities and other deeply overridden class hierarchies.  This class should never need any derived classes, so it's never needed.

gradle.properties
marchermans: What do you mean with "Minecolonies does not do that"?

uecasm: It does not mark the dependency as changing, nor does it include the extra code needed to make gradle not cache changing dependencies regardless (which is a new thing in 1.18 gradle, it appears).

Having said that, I've reverted this change for now (since it's not directly related to the rest of the PR), but I still strongly believe that it is the more correct solution and it should still be used.

uecasm: See [the gradle docs](https://docs.gradle.org/current/userguide/dynamic_versions.html#:~:text=By%20default%2C%20Gradle%20caches%20changing,resolve%20new%20snapshot%20versions%20earlier.) -- by default even "changing" deps are cached for 24 hours, which means that it will completely ignore your subsequent changes in the same day even if you republish (unless you remember to change the dep version on both sides every single time, which is a massive hassle).

Neither Minecolonies nor OPC includes the code to reduce the cache time.  And in my testing, reducing the caching time (even to a few seconds) is not completely reliable -- it will still sometimes use the cached version anyway.  Only using the snapshot tag will reliably always use the latest code.

(And even if that did work reliably, adding the changing tag to a dep is more obscure than just changing from a published version to a snapshot version -- especially as these settings are in different files -- so it's easier to miss and will cause more frustration when forgotten.)

Build scripts should be designed by default to make the lives of developers easier, not to make automated builds easier.  Automated builds can do more complex things to work around defaults; developers will usually forget.

uecasm:
Should I move this to 1.18.2?  Or what's the hold-up on merging it?

Raycoms:
Will use the weekend for another review, didn't have the time/muse

src/main/java/com/ldtteam/structurize/helpers/Settings.java
```diff
@@ -348,8 +353,7 @@ public void setActiveSchematic(final Blueprint blueprint)
     public void reset()
     {
         resetBlueprint();
-        rotation = 0;
-        isMirrored = false;
+        placement = new PlacementSettings();
```
Raycoms, line 370: Personally, I'd put the Wall stuff into a "WallSettings" which you can get from here, makes this much less cluttered.

src/main/java/com/ldtteam/structurize/management/StructureName.java
```diff
@@ -17,6 +21,9 @@
 {
     public static List<String> HUTS = Collections.synchronizedList(new ArrayList<>());
 
+    private static Map<String, Component> SECTIONS = createDefaultSections();
```
Raycoms, line 24: What is all of this for?

Raycoms (dismissed):

src/main/java/com/ldtteam/structurize/helpers/Settings.java
uecasm: `blueprint` is the original un-extended blueprint.  It is needed as a base to regenerate the extension and for network transfer.  `wall` is the extended blueprint (or exactly the same as `blueprint` when not extended), which is used in most other cases, e.g. for rendering and actual construction.

src/main/java/com/ldtteam/structurize/helpers/Settings.java
uecasm: That is already the case.  `PlacementSettings` holds a `WallExtents`.

Part of this PR actively reduces clutter by using `PlacementSettings` instead of the separate rotation/mirror (which also allows for future improvement to properly execute the mirroring by axis).

src/main/java/com/ldtteam/structurize/management/StructureName.java
uecasm: Custom build sections.  See examples in the Minecolonies PR.

uecasm:
Reminder that this is still waiting to exist.

Nightenom (commented):
Didn't look into depth but the approach of "baking" new blueprint using wallextents seems rather intrusive. As far as I understand this can be achieved simply by rendering the same blueprint multiple times and sending special multi-build message to server. Generally speaking baking new blueprint is rather fishy and I would avoid it as much as possible

Nightenom (commented):
Speaking about PlacementSettings I would rather see that as standalone pr so u can test it properly - it touches many core places

src/main/java/com/ldtteam/structurize/helpers/WallExtents.java
```diff
@@ -0,0 +1,152 @@
...
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.FriendlyByteBuf;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A description/config for wall replication.
+ */
+public final class WallExtents
+{
+    private short negative;
```
Nightenom, line 12: probably no point in using short

src/main/java/com/ldtteam/structurize/management/Structures.java
```diff
@@ -106,6 +107,12 @@ private Structures()
      */
     public static void init()
     {
+        md5Map.clear();
```
Nightenom, line 110: this looks weird - elaborate why is this needed

uecasm:
> Didn't look into depth but the approach of "baking" new blueprint using wallextents seems rather intrusive. As far as I understand this can be achieved simply by rendering the same blueprint multiple times and sending special multi-build message to server. Generally speaking baking new blueprint is rather fishy and I would avoid it as much as possible

The blueprints are not quite exact copies tiled; if the anchor block is a blueprint provider (e.g. a deco controller) it will only put one copy in; with modified TE data such that it is aware that it was tiled and it will upgrade as a single unit.  (This would in principle also avoid duplicating mcol hut blocks, except those are excluded from extending anyway.)  The extended blueprint is generated separately on client and server sides and is never transferred in full over the network, so it's "free" other than in memory.

> Speaking about PlacementSettings I would rather see that as standalone pr so u can test it properly - it touches many core places

Sure, I could probably split that out -- if it's likely to take less than months to review it again... 😬  But I have tested it very thoroughly as is.

src/main/java/com/ldtteam/structurize/management/Structures.java
uecasm: In testing, I noticed that deleting blueprint files and reloading the world, or switching between different worlds with different sets of blueprints in the same MC client session, did not properly reset the list of available blueprints and it got very confused.  This fixes that.

uecasm: It was also very noticeable when switching between different worlds with different section configs.

Nightenom:
> The blueprints are not quite exact copies tiled; if the anchor block is a blueprint provider (e.g. a deco controller) it will only put one copy in; with modified TE data such that it is aware that it was tiled and it will upgrade as a single unit. (This would in principle also avoid duplicating mcol hut blocks, except those are excluded from extending anyway.) The extended blueprint is generated separately on client and server sides and is never transferred in full over the network, so it's "free" other than in memory.

If it only ever exists on client then I really see no point in doing that, the only thing u basically do on client is moving preview copy and for that you can simply do "render at multiple poses". I'm actually asking what's the reason you need to "waste time" for creating such thing?
For the server "one deco controller to rule them all" I would rather see it as this:
- don't change blueprint on structurize size at all - since it's mcol thing anyway
- in many other games I saw, building a row of same type creates one build order per one type - as it's easier to maintain it :D
- add thing (on mcol side) to upgrade/fix multiple deco buildings of same type at once - probably behind research that can be leveled like: 1lvl walls - 2lvl parks or whatever we got there - 3lvl etc. and so on

> Sure, I could probably split that out -- if it's likely to take less than months to review it again... 😬 But I have tested it very thoroughly as is.

Time it takes depends on complexity of change and time I have (xd) should be better now as circus at other project got sane

uecasm:
> If it only ever exists on client

No, it exists on both, as I said.  Both need it for different reasons, but it avoids sending it over the network.

> "render at multiple poses".

No, because the copies are not identical.  They can have omitted blocks, both due to the "only one anchor" and because you can set an overlap.  It sounds like you're essentially proposing rebuilding it on the fly during each render tick, which seems much worse.

> don't change blueprint on structurize size at all

The code belongs in structurize because it is manipulating the internal Blueprint format.  Minecolonies should neither know nor care what that format is, nor should it try to manipulate it externally because that's making the interface between the two even more fragile than it already is.

> in many other games I saw, building a row of same type creates one build order per one type

That is an alternative implementation, yes.  It's not the one I wanted, and it's not really practical due to the possibility of overlapped schematics.

uecasm:
If it helps, you can think of what it's currently doing as a cache.  It sends the original unextended blueprint (if needed) plus the extents (the instructions to extend it) over the network.  Then it builds a cached blueprint once, and renders it as many times as needed from that cache.

This is not that dissimilar from what you were suggesting, except that it's better because it's cached and because it can handle the special cases of overlaps and anchors in a simpler fashion.

Nightenom:
You've completely misunderstood what I thought (mainly cuz Im lazy to type) will get to this later

Nightenom:
Sorry for staling this out, but I and Ray basically agreed on that we're fine with "spam placing" as long as it does not change the original blueprint, and it does not have to be tightened to walls only, and basically could be generally something like "grid placer"
Closing this because of old branch/old logic

Issue #489, opened 2022-02-28
Fix color box rendering


Nightenom:
Color boxes (like bounding boxed for scan/build tool or red boxes for anchors) do not work with certain shaders.
Current state: partial fix was made but still boxes sometimes do not render at all

PR #506, opened 2022-05-15
Fixes to fake world

### LINKED ISSUE #504: Crash with Epic Fight

- [X] I have searched the existing issues

### Are you using the latest MineColonies Verison?

- [X] I am running the latest **alpha** version of MineColonies for my Minecraft version.
I am also running the latest (possible) versions of other mods that are part of my problem.

- [X] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/). Or I was sent from discord to open an issue here.

### Minecraft Version

1.18

### MineColonies Version

minecolonies-1.18.2-1.0.673-ALPHA

### Structurize Version

structurize-1.18.2-1.0.389-RELEASE

Forge 40.1.0

### Reproduction Steps

Hello, I'm a creator of Epic Fight and I've got a crash report that conflicts with MineColonies. the direct reason for the crash is this:

```
Caused by: java.lang.NullPointerException: Cannot invoke "net.minecraft.world.level.GameRules.getBoolean(net.minecraft.world.level.GameRules$Key)" because the return value of "net.minecraft.world.level.Level.getGameRules()" is null
```

So, I wonder if MineColonies creates a level instance that returns a null value for getGameRules() method. If there's a reason it should be a null value, I'll just check if it is null before executing my code. Thanks.

The original page for this report is here: https://github.com/Yesssssman/epicfightmod/issues/990

### Logs

https://gist.github.com/Yesssssman/b384facc0a6041618bb66b3f293839ad

### Anything else?

_No response_

### Footer

---
## Viewers

Nightenom:
Closes #504 
Progress towards #489 

- bump gradle and forge versions (min forge is recommended build)
- update fake world and chunk classes
- make substitution blocks smarter in c6155de8efa844e44afd801eca7b8c7b0c163d91
- preview will now respect surrounding biomes in full update (after rotate/mirror change)
- general cleanup removing most warnings (just 2 deprecations left)

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
+++ b/src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
@@ -140,7 +140,7 @@ public class CreativeStructureHandler extends AbstractStructureHandler
     @Override
     public BlockState getSolidBlockForPos(final BlockPos worldPos)
     {
-        return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos);
+        return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos, null);
     }
 
     @Override
--- a/src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
+++ b/src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
@@ -142,4 +142,10 @@ public class CreativeStructureHandler extends AbstractStructureHandler
     {
         return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos);
     }
+
+    @Override
+    public BlockState getWorldGenBlock(final BlockPos worldPos, final BlockState virtualBlockAbove)
+    {
+        return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos, virtualBlockAbove);
+    }
 }
--- a/src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
+++ b/src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
@@ -144,7 +144,7 @@ public class CreativeStructureHandler extends AbstractStructureHandler
     }
 
     @Override
-    public BlockState getWorldGenBlock(final BlockPos worldPos, final BlockState virtualBlockAbove)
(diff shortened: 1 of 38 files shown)


src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
```diff
@@ -140,6 +140,12 @@ public void prePlacementLogic(final BlockPos worldPos, final BlockState blockSta
...
     {
-        return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos);
+        return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos, null);
+    }
+
+    @Override
+    public BlockState getWorldGenBlock(final BlockPos worldPos, final BlockState virtualBlockAbove)
+    {
+        return BlockUtils.getSubstitutionBlockAtWorld(getWorld(), worldPos, virtualBlockAbove);
     }
```
Raycoms, line 150: I don't get why that has to be an extra method. That should just be called by the default getSolidBlockForPos behaviour, but not have to exist in all handlers.

src/main/java/com/ldtteam/structurize/placement/structure/IStructureHandler.java
```diff
@@ -252,5 +252,14 @@ default BlockPos getProgressPosInWorld(final BlockPos localPos)
      * @param worldPos the world pos.
      * @return the right block (classically biome dependent).
      */
+    @Deprecated(forRemoval = true, since = "1.18.2")
     BlockState getSolidBlockForPos(BlockPos worldPos);
```
Raycoms, line 256: I would keep this, so that we can still have "a block we want" and not "only worldgen block"

src/main/java/com/ldtteam/structurize/placement/structure/CreativeStructureHandler.java
Nightenom: name addressed, leaving the method where it is - not gonna touch tape

src/main/java/com/ldtteam/structurize/placement/structure/IStructureHandler.java
Nightenom: name addressed

Nightenom (commented):
tested sp and mp

Raycoms (dismissed):

PR #508, opened 2022-05-22
Fix durability bar rendering and item count z-layering

### LINKED ISSUE #507: Blockout renders item icons with durability bars weirdly

- [X] I have searched the existing issues.

- [X] I am running the latest **alpha** version of Structurize for my Minecraft version.

- [X] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/).

### Minecraft Version

1.16

### Structurize Version

1.0.392-RELEASE

_No response_

_No response_

### Current Behavior

image

The durability bar appears to be rendered almost full height.

### Expected Behavior

The durability bar should only be at the bottom of the icon, same as vanilla.

### Reproduction Steps

1. Put some items with non-full durability in the Warehouse.
2. Click on the search button.

### Logs

n/a

### Anything else?

It displays as expected in 1.18.

### Footer

---
## Viewers

uecasm:
Closes #507

- Fixes durability bar rendering
- Fixes item counts layering above tooltips
- Makes the item rendering use the named constant a bit more so that the value derivation is more obvious

Review please

This is 1.16 only; 1.18 appears to render these things properly (though I didn't inspect the code to see if it might benefit from some of the other changes).

(FWIW, it appears that the main issue with durability bar rendering is that the parameters to `fill` are right and bottom, not width and height.)

### DIFF
--- a/src/main/java/com/ldtteam/blockout/controls/ItemIcon.java
+++ b/src/main/java/com/ldtteam/blockout/controls/ItemIcon.java
@@ -27,7 +27,8 @@ import net.minecraftforge.registries.ForgeRegistries;
  */
 public class ItemIcon extends Pane
 {
-    private static final float DEFAULT_ITEMSTACK_SIZE = 16f;
+    private static final int DEFAULT_ITEMSTACK_SIZE = 16;
+    private static final float DEFAULT_ITEMSTACK_SIZEF = DEFAULT_ITEMSTACK_SIZE;
 
     /**
      * ItemStack represented in the itemIcon.
@@ -93,7 +94,7 @@ public class ItemIcon extends Pane
         {
             ms.pushPose();
             ms.translate(x, y, 0.0f);
-            ms.scale(this.getWidth() / DEFAULT_ITEMSTACK_SIZE, this.getHeight() / DEFAULT_ITEMSTACK_SIZE, 1.0f);
+            ms.scale(this.getWidth() / DEFAULT_ITEMSTACK_SIZEF, this.getHeight() / DEFAULT_ITEMSTACK_SIZEF, 1.0f);
 
             FontRenderer font = itemStack.getItem().getFontRenderer(itemStack);
             if (font == null)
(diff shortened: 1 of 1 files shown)


src/main/java/com/ldtteam/blockout/controls/ItemIcon.java
```diff
@@ -122,11 +123,11 @@ private void renderGuiItemDecorations(
         if (stack.getCount() != 1)
         {
             String s = String.valueOf(stack.getCount());
-            matrixstack.translate(0.0D, 0.0D, 200.0D);
+            matrixstack.translate(0.0D, 0.0D, 150.0D);
             IRenderTypeBuffer.Impl irendertypebuffer$impl = IRenderTypeBuffer.immediate(Tessellator.getInstance().getBuilder());
             fontRenderer.drawInBatch(s,
-              (float) (19 - 2 - fontRenderer.width(s)),
-              (float) (6 + 3),
+              DEFAULT_ITEMSTACK_SIZEF + 1 - fontRenderer.width(s),
```
Nightenom, line 129: Please refrain from constanting vanilla inlines

Nightenom (commented):
One minor, otherwise mergeable
Btw since it's inline there should be different reason why it does not work: either vanilla updated or I inlined wrong arguments
And iirc 1.18 doesn't have this inline anymore

src/main/java/com/ldtteam/blockout/controls/ItemIcon.java
uecasm: What do you mean by that?

Nightenom: This code is vanilla inline = code from vanilla which can't be for some reason called directly (usually code change, same thing as minecart/fishing rod in mcol)
Thus it doesn't make sense to put constants in it since you need to changes as easier as possible

uecasm: But (a) you're not going to change it again, as 1.18 has already moved on, and (b) it makes sense to use the constants where they're meaningful, otherwise why have them? and (c) how was I supposed to know that?  There's nothing in the code that indicates that.

Nightenom: a) true
b) true except this since then its harder to compare with vanilla
c) true, its probably that old it does not have comment with inline description, but u could guess that from decompiler-codestyle

Nightenom:
Closed in 5375595
Closes #507

PR #510, opened 2022-05-30
Add an entity icon control

uecasm:
- Adds an `<entityicon>` control to blockout
- Uses that control in the scan tool UI
    image

Review please

Items are a bit wibbly, and don't stay in their lane, but I don't think that's entirely a bad thing.
The entity rendering code is based on the JEI worker rendering from MineColonies, but with more auto-scaling and support for more entity types.


Raycoms (approved):
Loving it, please also PR to blockUI

marchermans (approved):
This looks very nice!

PR #515, opened 2022-06-29
New backend

Raycoms:
This is how the new backend will look like.

In a nutshell:

- We have structure packs that we sync ourselves between client and server.
- All IO is done off-thread
- Server updates client with missing/outdated packs
- Client and server index pack meta data on startup of all the packs they hold and compare them
- All done with NIO

The goal is to have a system that is:
a) As simple as possible
b) Works entirely off-thread
c) Doesn't depend on any Minecraft loading order things
d) Depends as little on mojang as possible.
As the one doing most of the porting, this matters =P

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/network/messages/NotifyClientAboutStructurePacks.java
+++ b/src/main/java/com/ldtteam/structurize/network/messages/NotifyClientAboutStructurePacks.java
@@ -1,0 +1,77 @@
+package com.ldtteam.structurize.network.messages;
+
+import com.ldtteam.structurize.storage.ClientStructurePackLoader;
+import com.ldtteam.structurize.storage.StructurePackMeta;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Notify the client about the structure packs on the server side.
+ */
+public class NotifyClientAboutStructurePacks implements IMessage
+{
+    /**
+     * List of server structure packs.
+     * Contains String Name, and Integer version.
+     */
+    final Map<String, Integer> serverStructurePacks = new HashMap<>();
+
+    /**
+     * Public standard constructor.
+     */
+    public NotifyClientAboutStructurePacks(final FriendlyByteBuf buf)
+    {
+        final int length = buf.readInt();
+        for (int i = 0; i < length; i++)
+        {
+            this.serverStructurePacks.put(buf.readUtf(32767), buf.readInt());
+        }
+    }
+
+    /**
+     * Notify the client about the server structurepacks.
+     * @param clientStructurePacks the list of packs.
+     */
+    public NotifyClientAboutStructurePacks(final Map<String, StructurePackMeta> clientStructurePacks)
+    {
+        for (final StructurePackMeta pack : clientStructurePacks.values())
+        {
+            if (!pack.isImmutable())
+            {
+                this.serverStructurePacks.put(pack.getName(), pack.getVersion());
+            }
+        }
+    }
+
+    @Override
+    public void toBytes(final FriendlyByteBuf buf)
+    {
+        buf.writeInt(this.serverStructurePacks.size());
+        for (final Map.Entry<String, Integer> packInfo : this.serverStructurePacks.entrySet())
+        {
+            buf.writeUtf(packInfo.getKey());
+            buf.writeInt(packInfo.getValue());
+        }
+    }
+
+    @Nullable
+    @Override
+    public LogicalSide getExecutionSide()
+    {
+        return LogicalSide.CLIENT;
+    }
+
+    @Override
+    public void onExecute(final NetworkEvent.Context ctxIn, final boolean isLogicalServer)
+    {
+        if (!isLogicalServer)
+        {
+            ClientStructurePackLoader.onServerSyncAttempt(this.serverStructurePacks);
+        }
+    }
+}
--- a/src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
+++ b/src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
@@ -1,0 +1,274 @@
+package com.ldtteam.structurize.storage;
+
+import com.ldtteam.structurize.Network;
+import com.ldtteam.structurize.Structurize;
+import com.ldtteam.structurize.api.util.Log;
+import com.ldtteam.structurize.network.messages.NotifyServerAboutStructurePacks;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufInputStream;
+import net.minecraft.Util;
+import net.minecraft.client.Minecraft;
+import net.minecraftforge.event.TickEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.ModList;
+import net.minecraftforge.forgespi.language.IModInfo;
+import org.codehaus.plexus.util.FileUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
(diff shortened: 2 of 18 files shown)


uecasm:
It would be nice if it also supported loading schematic packs from datapacks (one datapack might contain one or more schematic packs).  They're already a nice way to distribute files (or keep them loose while in development), and there's already some built-in synching that might be sufficient?  Unless there's some technical constraint (size limits etc)?

Raycoms:
> It would be nice if it also supported loading schematic packs from datapacks (one datapack might contain one or more schematic packs). They're already a nice way to distribute files (or keep them loose while in development), and there's already some built-in synching that might be sufficient? Unless there's some technical constraint (size limits etc)?

Yeah, I don't want to support datapacks because they're synched on login to a server. While there are technically no size constraints. In practice, this results in long login delays. Thats why, on purpose, I moved that "off-pack"

uecasm:
> Yeah, I don't want to support datapacks because they're synched on login to a server. While there are technically no size constraints. In practice, this results in long login delays. Thats why, on purpose, I moved that "off-pack"

If the packs persist across multiple sessions and servers (and support having multiple versions of the "same" pack simultaneously, for servers that didn't upgrade at the same time), that might be less of an issue.  You'd have to download them the first time you connect and if there's any update, but most of the time logins would be fast.  Though yeah, background loading is nicer in general (until you want to build something as soon as you log in and it's not there...)

uecasm:
Actually, wait a tick.  *Data*packs don't get synched from server to client AFAIK, they remain purely server-side unless you do something explicitly to sync them (e.g. recipes and tags have explicit synching, or you can do your own like we do with custom crafter recipes).  So you could still support loading schematic packs from datapacks on the server side (or in singleplayer), and then asynchronously sync them as individual schematic packs via your proposed mechanism regardless of where they were loaded from.

It's only a *resource* pack that (optionally) gets synched as a pack on login.  But schematics should not be resource packs anyway (despite currently living in the assets folder, which is the wrong place).

Raycoms:
> Actually, wait a tick. _Data_packs don't get synched from server to client AFAIK, they remain purely server-side unless you do something explicitly to sync them (e.g. recipes and tags have explicit synching, or you can do your own like we do with custom crafter recipes). So you could still support loading schematic packs from datapacks on the server side (or in singleplayer), and then asynchronously sync them as individual schematic packs via your proposed mechanism regardless of where they were loaded from.
> 
> It's only a _resource_ pack that (optionally) gets synched as a pack on login. But schematics should not be resource packs anyway (despite currently living in the assets folder, which is the wrong place).

I put it on purpose outside of "mojangs reach" to avoid any cross interaction, necessity of porting, potential synching, etc.
Datapacks are also "per-world" which is also a pain.

uecasm:
> Datapacks are also "per-world" which is also a pain.

That's a feature, not a bug.  Different servers can have different custom schematic packs.  Mods and modpacks can provide datapacks that are loaded by default into new worlds, and the worlds can selectively disable some of them (e.g. a medieval themed server might want to disable some of the non-medieval styles).  And you could still have a global folder for client-side custom packs/decos.

Raycoms:
> > Datapacks are also "per-world" which is also a pain.
> 
> That's a feature, not a bug. Different servers can have different custom schematic packs. Mods and modpacks can provide datapacks that are loaded by default into new worlds, and the worlds can selectively disable some of them (e.g. a medieval themed server might want to disable some of the non-medieval styles).

That is all fine, the client won't see the disabled ones on the server.

Raycoms:
> > > Datapacks are also "per-world" which is also a pain.
> > 
> > 
> > That's a feature, not a bug. Different servers can have different custom schematic packs. Mods and modpacks can provide datapacks that are loaded by default into new worlds, and the worlds can selectively disable some of them (e.g. a medieval themed server might want to disable some of the non-medieval styles).
> 
> That is all fine, the client won't see the disabled ones on the server.

These packs are much simpler than datapacks. You just put them in the root in the "blueprints" folder. That's it. Done

Raycoms:
> > > > Datapacks are also "per-world" which is also a pain.
> > > 
> > > 
> > > That's a feature, not a bug. Different servers can have different custom schematic packs. Mods and modpacks can provide datapacks that are loaded by default into new worlds, and the worlds can selectively disable some of them (e.g. a medieval themed server might want to disable some of the non-medieval styles).
> > 
> > 
> > That is all fine, the client won't see the disabled ones on the server.
> 
> These packs are much simpler than datapacks. You just put them in the root in the "blueprints" folder. That's it. Done

And, like I said, this way we don't depend on any mojang datapack loading oddities or similar stuff

marchermans (changes requested):
I agree mostly with @uecasm 
I would like to see a way more flexible API design here.

Where loaders can pick and chose, on a blueprint or category level what the metadata returned is, not on pack level. Mostly because the bandwidth costs are way higher if the user has to download the entire pack with possibly hundreds of blueprints from a remote server which is way more bandwidth then needed.

Additionally the complete refusal to create any system which can either 1) use vanilla datapacks or 2) which modders can properly extend and use themselves makes this a bit like the old system. Completely closed of, made to tailor the needs of minecolonies and not as if this is a standalone mod.

Raycoms:
> I agree mostly with @uecasm I would like to see a way more flexible API design here.
> 
> Where loaders can pick and chose, on a blueprint or category level what the metadata returned is, not on pack level. Mostly because the bandwidth costs are way higher if the user has to download the entire pack with possibly hundreds of blueprints from a remote server which is way more bandwidth then needed.
> 
> Additionally the complete refusal to create any system which can either 1) use vanilla datapacks or 2) which modders can properly extend and use themselves makes this a bit like the old system. Completely closed of, made to tailor the needs of minecolonies and not as if this is a standalone mod.

Like I said, if a modder wants datapack support, they can add an additional loader for the client/server to add things to it.
Our system is very similar to datapacks I agree. But also different enough that I do not want to use it.
Our packs are instance global, are indexed dynamically on our terms (i.e. on usage), are synched on our own terms, etc.

src/main/java/com/ldtteam/structurize/network/messages/NotifyClientAboutStructurePacks.java
```diff
@@ -0,0 +1,77 @@
...
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Notify the client about the structure packs on the server side.
+ */
+public class NotifyClientAboutStructurePacks implements IMessage
```
someaddons, line 16: ClientStructurePacksMessage ?

Raycoms: No, it notifies the client, so would more be of a ServerStructurePacksMessage

someaddons: I'd see client in name as a message dedicated for the client to receive^^

src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
```diff
@@ -0,0 +1,274 @@
...
+        {
+            if (Minecraft.getInstance().level != null && Minecraft.getInstance().level.getGameTime() % 20 == 0 && loadingState == ClientLoadingState.FINISHED_LOADING)
+            {
+                loadingState = ClientLoadingState.SYNCING;
+                Network.getNetwork().sendToServer(new NotifyServerAboutStructurePacks(StructurePacks.packMetas));
+            }
+            else if (Minecraft.getInstance().level == null && (loadingState == ClientLoadingState.SYNCING || loadingState == ClientLoadingState.FINISHED_SYNCING))
+            {
+                Log.getLogger().warn("Client logged off. Resetting Pack Meta and Reloading State");
+                loadingState = ClientLoadingState.LOADING;
```
someaddons, line 115: shouldn't this be `ClientLoadingState.FINISHED_LOADING` else we're reverting to the state we had before the client started

Raycoms: Yes, this resets the loading entirely and starts over. The next call basically sets this to finished loading then

someaddons: I'm not seeing where it starts over then, since both conditions here require different loading states

Raycoms: onClientStarting();

someaddons (approved):

PR #517, opened 2022-07-14
Structurize 1.5

Raycoms:
Closes #
Closes #
Closes #

This has the all the adjustments for the new structurize backend.
Could almost be called half way of a structurize 2.0

- Loading structures more flexibly via structure packs
- Syncing between client and server
- Additional metadata in blueprints
- Support to render multiple blueprints at the same time
- Async handling of blueprints.

A few details:

The API is the StructurePacks file, where blueprints can be Stored and retrieved from given a structure pack id (which leads to a structure pack) and the path within the structure pack. Structure pack metadata allows to locate the root of the structure pack folder. 

The client and server loaders are mostly for client/server synching on startup and to initially discover structure packs. However, mods can also dynamically adjust their packs whenever they want.

Structure packs are stored in the mod resources but neither in the datapacks nor in the assets to avoid confusion. We handle syncing ourselves and treat the packs as "instance global". Similarly, locally players can drop them off in the root instance folder in a "blueprints" folder.

On scan, a special player folder is created inside this blueprints folder for the scans.
Client side scans are also synced in a special way to the server to avoid servers having to distribute all of those to the other clients.

$todo: Readjust the properties file

Review please


src/main/java/com/ldtteam/structurize/blockentities/interfaces/ISpecialCreativeHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import com.ldtteam.structurize.placement.structure.AbstractStructureHandler;
+import com.ldtteam.structurize.util.PlacementSettings;
+import net.minecraft.core.BlockPos;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.level.Level;
+
+/**
+ * Special creative placement handler anchors.
+ */
+public interface ISpecialCreativeHandler
```
marchermans: Change the name to indicate that this needs to be implemented on an anchor block.

src/main/java/com/ldtteam/structurize/blocks/interfaces/IBlueprintDataProvider.java
```diff
@@ -24,6 +24,8 @@
     public final static String TAG_TAG_NAME_LIST  = "tagNameList";
```
marchermans, line 24: Rename this interface to make it clear that it is supposed to be implemented on the block entity of the anchor.

marchermans: Possibly move to different package.

src/main/java/com/ldtteam/structurize/client/gui/AbstractBlueprintManipulationWindow.java
```diff
@@ -0,0 +1,365 @@
...
+        if (openTicks++ >= 20 * 10)
+        {
+            findPaneOfTypeByID("tip", Text.class).setVisible(false);
+        }
+    }
+
+    @Override
+    public boolean onKeyTyped(final char ch, final int key)
+    {
+        if (ch == '\u0000')
```
marchermans, line 120: Handle implementation of the mirrot button here somewhere as well.

marchermans: Move this logic to an option in BlockUI. Possibly somehow linked to keybinds in the key binding menu? Not sure how that works.

src/main/java/com/ldtteam/structurize/network/NetworkChannel.java
```diff
@@ -11,25 +17,53 @@
...
+
+    /**
+     * The message that this channel can process, as viewed from a message type.
+     */
+    private final Map<Class<? extends IMessage>, Integer> messageTypeToIdMap = Maps.newHashMap();
+
+    /**
+     * Cache of partially received messages, this holds the data untill it is processed.
+     */
+    private final Cache<Integer, Map<Integer, byte[]>> messageCache = CacheBuilder.newBuilder()
```
marchermans, line 57: If you want you might want to fix this here, since on the server side this could collide if players send two packets with the same com id.

src/main/java/com/ldtteam/structurize/network/messages/ClientBlueprintRequestMessage.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Request a blueprint from the client.
+ */
+public class ClientBlueprintRequestMessage implements IMessage
+{
+    /**
+     * List of pending blueprint requests.
+     */
+    public static final Queue<Tuple<ClientBlueprintRequestMessage, Future<byte[]>>> blueprintRequestQueue = new LinkedList<>();
```
marchermans: No. This is not part of the message, move this to some dedicated mechanism or handler.

src/main/java/com/ldtteam/structurize/storage/BlueprintPlacementHandling.java
```diff
@@ -0,0 +1,155 @@
...
+        if (!StructurePacks.packMetas.containsKey(message.structurePackId))
+        {
+            final Path blueprintPath = new File(".").toPath()
+              .resolve(BLUEPRINT_FOLDER)
+              .resolve(CLIENT_FOLDER)
+              .resolve(message.player.getUUID().toString())
+              .resolve(message.structurePackId)
+              .resolve(message.blueprintPath);
+
+            ServerBlueprintFutureProcessor.consumerQueue.add(new ServerBlueprintFutureProcessor.ProcessingData(StructurePacks.getBlueprintFuture(message.structurePackId, blueprintPath),
```
marchermans: Don't expose the consumer queue directly, expose its intended use via methods.

src/main/java/com/ldtteam/structurize/storage/ClientBlueprintFutureProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Waits for blueprint futures to finish loading and then processes them.
+ */
+public class ClientBlueprintFutureProcessor
+{
+    /**
+     * Queue for processing.
+     */
+    public static final Queue<ProcessingData> consumerQueue = new LinkedList<>();
```
marchermans: Don't expose fields, expose their use case in a method, so don't just add a getter.

src/main/java/com/ldtteam/structurize/storage/ServerBlueprintFutureProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Waits for blueprint futures to finish loading and then processes them.
+ */
+public class ServerBlueprintFutureProcessor
+{
+    /**
+     * Queue for processing.
+     */
+    public static final Queue<ProcessingData> consumerQueue = new LinkedList<>();
```
marchermans: See other places where you expose fields.

src/main/java/com/ldtteam/structurize/storage/StructurePacks.java
```diff
@@ -0,0 +1,614 @@
...
+
+    /**
+     * Get a blueprint future.
+     * @param structurePackId the structure pack the blueprint is in.
+     * @param subPath the path of the specific blueprint in the pack.
+     * @return the blueprint future (might contain null).
+     */
+    public static Future<Blueprint> getBlueprintFuture(final String structurePackId, final String subPath)
+    {
+        return Util.ioPool().submit(() -> getBlueprint(structurePackId, subPath));
```
marchermans: Lets just use our own pool here.
I think that is more resilient for the future.

src/main/java/com/ldtteam/structurize/storage/SurvivalBlueprintHandlers.java
```diff
@@ -0,0 +1,68 @@
...
+        return handlers.get(id);
+    }
+
+    /**
+     * Getthe full list of handlers.
+     * @return a list.
+     */
+    public static List<ISurvivalBlueprintHandler> getHandlers()
+    {
+        return new ArrayList<>(handlers.values());
```
marchermans: Make this an unmodifyable list, store that unmodifyable view as a field. Makes calls to this way cheaper.

src/main/java/com/ldtteam/structurize/storage/rendering/RenderingCache.java
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * Boxes to render.
+     */
+    public static Map<String, ScanPreviewData> boxRenderingCache = new HashMap<>();
+
+    /**
+     * Blueprints to render.
+     */
+    public static Map<String, BlueprintPreviewData> blueprintRenderingCache = new HashMap<>();
```
marchermans: Public fields.....

src/main/java/com/ldtteam/structurize/storage/rendering/ServerPreviewDistributor.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Class handling blueprint syncing between players.
+ */
+public class ServerPreviewDistributor
+{
+    /**
+     * Players that signed up to receive blueprint data.
+     */
+    public static Map<UUID, Tuple<ServerPlayer, BlueprintRenderSettings>> registeredPlayers = new HashMap<>();
```
marchermans: Public field

src/main/java/com/ldtteam/structurize/storage/rendering/types/BlueprintPreviewData.java
```diff
@@ -0,0 +1,288 @@
...
+public class BlueprintPreviewData
+{
+    @NotNull
+    public Rotation rotation = Rotation.NONE;
+
+    @NotNull
+    public Mirror mirror = Mirror.NONE;
+
+    @Nullable
+    public BlockPos pos;
```
marchermans: public field.

src/main/java/com/ldtteam/structurize/storage/rendering/types/ScanPreviewData.java
```diff
@@ -0,0 +1,29 @@
...
+public class ScanPreviewData
+{
+    @NotNull
+    public final BlockPos pos1;
+
+    @NotNull
+    public final BlockPos pos2;
+
+    @NotNull
+    public Optional<BlockPos> anchor;
```
marchermans: public fields.

marchermans (dismissed):
See the comments.

src/api/java/com/ldtteam/structurize/api/util/Utils.java
```diff
@@ -14,6 +22,44 @@ private Utils()
     {
     }
 
+    /**
+     * Play a success sound.
+     * @param player the player to play it for.
+     * @param position the position it is played at.
+     */
+    public static void playSuccessSound(@NotNull final Player player, @NotNull final BlockPos position)
```
Nightenom, line 30: Honestly since its logic based method I would use serverplayer for argument to mitigate client usage (as much as possible)

src/main/java/com/ldtteam/structurize/Structurize.java
```diff
@@ -52,9 +64,26 @@ public Structurize()
         ModBlockEntities.getRegistry().register(FMLJavaModLoadingContext.get().getModEventBus());
         Mod.EventBusSubscriber.Bus.MOD.bus().get().register(LifecycleSubscriber.class);
         Mod.EventBusSubscriber.Bus.FORGE.bus().get().register(EventSubscriber.class);
+
+        DistExecutor.unsafeRunWhenOn(Dist.CLIENT, () -> ClientStructurePackLoader::onClientLoading);
+        DistExecutor.unsafeRunWhenOn(Dist.DEDICATED_SERVER,  () -> ServerStructurePackLoader::onServerStarting);
+
+        Mod.EventBusSubscriber.Bus.FORGE.bus().get().register(ServerStructurePackLoader.class);
```
Nightenom, line 71: meh, weve got event handlers for this

src/main/java/com/ldtteam/structurize/Structurize.java
```diff
@@ -52,9 +64,26 @@ public Structurize()
         ModBlockEntities.getRegistry().register(FMLJavaModLoadingContext.get().getModEventBus());
         Mod.EventBusSubscriber.Bus.MOD.bus().get().register(LifecycleSubscriber.class);
         Mod.EventBusSubscriber.Bus.FORGE.bus().get().register(EventSubscriber.class);
+
+        DistExecutor.unsafeRunWhenOn(Dist.CLIENT, () -> ClientStructurePackLoader::onClientLoading);
+        DistExecutor.unsafeRunWhenOn(Dist.DEDICATED_SERVER,  () -> ServerStructurePackLoader::onServerStarting);
```
Nightenom, line 69: calling this in mod init is sus, methods should be named differently or called from sided setup events

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -305,15 +310,15 @@ else if (tileEntity instanceof EnchantmentTableBlockEntity enchTable)
                 else if (tileEntity instanceof CampfireBlockEntity campfire)
                 {
                     final BlockState bs = blockAccess.getBlockState(tePos);
-                    if (bs.getValue(CampfireBlock.LIT))
+                    if (bs.getBlock() instanceof CampfireBlock && bs.getValue(CampfireBlock.LIT))
```
Nightenom, line 313: This should not be necessary, since if there is BE and block is not matching, then u've problem somewhere else imho

src/main/java/com/ldtteam/structurize/client/gui/AbstractBlueprintManipulationWindow.java
Nightenom: Yeah, definitely needs keybindings logic. Also if nothing else then use GLFW.key_blah_bleh. Using char is horrible a buggy af decision

src/main/java/com/ldtteam/structurize/config/BlueprintRenderSettings.java
```diff
@@ -0,0 +1,67 @@
...
+{
+    /**
+     * Singleton instance of the settings for the client side.
+     */
+    public static BlueprintRenderSettings instance = new BlueprintRenderSettings();
+
+    /**
+     * All the settings.
+     */
+    public Map<String, Boolean> renderSettings = new HashMap<>();
```
Nightenom, line 21: String based map? is this public API? if so then still I would use iface with single string noarg method and use Enum in our space, otherwise it should be enum map

src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
```diff
@@ -1,34 +1,51 @@
...
 import net.minecraftforge.event.TickEvent.Phase;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
+
 import java.util.List;
 import java.util.Map;
 
 public class ClientEventSubscriber
 {
+    @SubscribeEvent
+    public static void renderWorldLastEvent(final RenderGameOverlayEvent.PreLayer event)
```
Nightenom, line 39: wrong method name

src/main/java/com/ldtteam/structurize/event/LifecycleSubscriber.java
```diff
@@ -22,7 +21,6 @@ public static void onModInit(final FMLCommonSetupEvent event)
     {
         Network.getNetwork().registerCommonMessages();
```
Nightenom, line 22: U have this in mod class, duplicated call

src/main/java/com/ldtteam/structurize/network/messages/SyncPreviewCacheToClient.java
```diff
@@ -0,0 +1,72 @@
...
+import com.ldtteam.structurize.storage.rendering.types.BlueprintPreviewData;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Sync blueprint preview data to the client.
+ */
+public class SyncPreviewCacheToClient implements IMessage
```
Nightenom, line 13: missing message in class name

src/main/java/com/ldtteam/structurize/network/messages/SyncPreviewCacheToServer.java
```diff
@@ -0,0 +1,54 @@
...
+import com.ldtteam.structurize.storage.rendering.types.BlueprintPreviewData;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Sync blueprint preview data to the server.
+ */
+public class SyncPreviewCacheToServer implements IMessage
```
Nightenom, line 13: missing message in class name

src/main/java/com/ldtteam/structurize/network/messages/SyncSettingsToServer.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.util.Tuple;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Sync player settings to server.
+ */
+public class SyncSettingsToServer implements IMessage
```
Nightenom, line 14: missing message in class name

src/main/java/com/ldtteam/structurize/network/messages/TransferStructurePackToClient.java
```diff
@@ -0,0 +1,78 @@
...
+import io.netty.buffer.Unpooled;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Transfer a zipped structure pack to the client.
+ */
+public class TransferStructurePackToClient implements IMessage
```
Nightenom, line 14: missing message in class name

src/main/java/com/ldtteam/structurize/proxy/IProxy.java
```diff
@@ -27,15 +28,4 @@ default void openBuildToolWindow(final BlockPos pos, final int groundstyle)
...
     {
     }
-
-    /**
-     * Get the file representation of the additional schematics' folder.
-     *
-     * @return the folder for the schematic
-     */
-    @Nullable
-    default File getSchematicsFolder()
```
Nightenom, line 37: Honestly these proxies should get removed

src/main/java/com/ldtteam/structurize/storage/StructurePacks.java
```diff
@@ -0,0 +1,697 @@
...
+     * @return the blueprint or null.
+     */
+    @Nullable
+    public static Blueprint getBlueprint(final String structurePackId, final String subPath)
+    {
+        while (!finishedLoading)
+        {
+            try
+            {
+                Thread.sleep(10);
```
Nightenom, line 356: wut face, can we get something more "legal" here than Thread#sleep ?
especially when this code is copied ten times, it would be great to have one method saying (if) why sleep

src/main/java/com/ldtteam/structurize/storage/StructurePackMeta.java
```diff
@@ -0,0 +1,213 @@
...
+            return subPath.toString().replace(rootPath + "/", "");
+        }
+    }
+
+    /**
+     * Get the normalized sub path from this pack.
+     * @param subPath the path we're in.
+     * @return the relative sub path.
+     */
+    public String getNormalizedSubPath(final String subPath)
```
Nightenom, line 202: sus method

src/main/java/com/ldtteam/structurize/storage/StructurePackMeta.java
```diff
@@ -0,0 +1,213 @@
...
+
+        this.mods = new ArrayList<>();
+        final JsonArray modsArray = json.get("mods").getAsJsonArray();
+        for (int i = 0; i < modsArray.size(); i++)
+        {
+            mods.add(modsArray.get(i).getAsString());
+        }
+        this.rootPath = rootPath;
+        this.version = json.get("version").getAsInt();
+        this.packFormat  = json.get("pack-format").getAsInt();
```
Nightenom, line 88: I dont like this being parsed/written in many places, ie. string constants being everywhere
(saw this like three times already)

src/main/java/com/ldtteam/structurize/util/IOPool.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Structurize specific thread factory.
+     */
+    public static class StructurizeThreadFactory implements ThreadFactory
+    {
+        /**
+         * Ongoing thread IDs.
+         */
+        public static int id;
```
Nightenom, line 31: while the executor is static this id being static is very sus, either non-static or atomic

src/main/java/com/ldtteam/structurize/util/IOPool.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Creates a new thread pool for pathfinding jobs
+     *
+     * @return the threadpool executor.
+     */
+    public static ThreadPoolExecutor getExecutor()
+    {
+        if (executor == null)
+        {
+            executor = new ThreadPoolExecutor(1, 2, 10, TimeUnit.SECONDS, ioQueue, new StructurizeThreadFactory());
```
Nightenom, line 53: the amount of threads should definitely be configurable, with zero meaning immediate on-same-thread execution

Nightenom (commented):
well, ive told u to ping me when mergeable not when merger :shrug:
generally ok but:
blueprint "coordinate" should be record atleast from pack and path, if ever online thingy then author will be probably imaginary root (not pack, to allow more ppl name the pack in same way)
i like the offthreading but its missing config entry (it should be common config, since server would override client value)
the way you redesigned syncing is regression, ppl should be able to filter out with whom they want to share to

I didn't look deeply into ui and the blueprint managment itself, will critize (xd) the ui later, hopefully u managed to redesign the management in way u wanted and tested it properly (would require me to run ide to actually review that in meaningful manner)

src/api/java/com/ldtteam/structurize/api/util/Utils.java
Raycoms: It's actually okay to call success sounds on the client too

src/main/java/com/ldtteam/structurize/Structurize.java
Raycoms: I know, but I wanted a bigger separation of concerns

src/main/java/com/ldtteam/structurize/storage/StructurePackMeta.java
Raycoms: It's for windows/linux server/client sync of subpaths

Raycoms:
> well, ive told u to ping me when mergeable not when merger shrug generally ok but: blueprint "coordinate" should be record atleast from pack and path, if ever online thingy then author will be probably imaginary root (not pack, to allow more ppl name the pack in same way) i like the offthreading but its missing config entry (it should be common config, since server would override client value) the way you redesigned syncing is regression, ppl should be able to filter out with whom they want to share to
> 
> I didn't look deeply into ui and the blueprint managment itself, will critize (xd) the ui later, hopefully u managed to redesign the management in way u wanted and tested it properly (would require me to run ide to actually review that in meaningful manner)

Yeah, there are certainly some things that can be extended upon.
On the syncing, this is on purpose to make this much much simpler for the user. The command stuff was very powerful but a big overkill. Maybe in the future we could make a "team setting" where people can set a team to share it with or so.
You can check out the Minecolonies part of the PR to see how we now hook into things

src/api/java/com/ldtteam/structurize/api/util/Utils.java
Nightenom: Yeah, but the method doesnt do that, so either should do both or somehow state that it's onesided

src/main/java/com/ldtteam/structurize/Structurize.java
Nightenom: still meh, you can do `onEvent(blah) { someRandomClass.doSomething(data from event); }` in event handler

src/main/java/com/ldtteam/structurize/storage/StructurePackMeta.java
Nightenom: like if it's used in blueprint coordinate then it should have fixed logic like ResLoc do, if not then I would expect nio Path being used, this really smells fishy

Nightenom:
> On the syncing, this is on purpose to make this much much simpler for the user. The command stuff was very powerful but a big overkill. Maybe in the future we could make a "team setting" where people can set a team to share it with or so.

I don't actually care about that you made it easier, it's just weird to do regression during update/rewrite

> You can check out the Minecolonies part of the PR to see how we now hook into things

Ain't got time to do that :D

src/main/java/com/ldtteam/structurize/storage/StructurePackMeta.java
Raycoms: Yeah, it's a pain, we sync it from client to server (and from server to client) so we can't use a path so we can't use NIO values but we can't really use res locations either because we might use invalid chars in it (afaik at least). And on the top of that in jar locations on windows are "/" and out of the jar in "\"

Nightenom: well I know syncing paths suck a lot, but that's why we should do stable system for it, and honestly meh invalid chars, I would go with same (or maybe a bit looser, but no acutes etc. and always lowercase) restrictions as in resLoc

Raycoms: but what if there are folders that don't respect that. So players would need to respect that with folder and file names too, which is a pita. I know that this is not perfect. But I never intended it to be perfect nor "100% finished". If I did 100% finished, I'd be done next year. But it doesn't have to be 100% finished to be better than the status quo, flexible enough to allow decent improvements and feature complete

PR #519, opened 2022-07-23
Fix tooltips going missing

uecasm:
- Fixes tooltips only occasionally working

Review please

The issue is that the array directly returned is cached (for a certain number of open windows), but is then also modified when the window is shown:
https://github.com/ldtteam/Structurize/blob/8e8cb55dca9a6b14ef9b80698170143e7e3a8b6a/src/main/java/com/ldtteam/blockout/Pane.java#L504

The net result is that the tooltip (when loaded from the xml, rather than being set programmatically) works exactly once on first load, then doesn't work again until either an F3+T reload is triggered or enough other windows are shown that it gets evicted from the cache and has to reload from XML.

This bug will likely affect BlockUI as well, since it has the same code.  Although I have not recently retested this and I do vaguely recall it behaving better the last time I did, but perhaps I was overlooking the cache effect.

### DIFF
--- a/src/main/java/com/ldtteam/blockout/Pane.java
+++ b/src/main/java/com/ldtteam/blockout/Pane.java
@@ -91,7 +91,7 @@ public class Pane extends AbstractGui
         visible = params.getBoolean("visible", visible);
         enabled = params.getBoolean("enabled", enabled);
         onHoverId = params.getString("onHoverId", onHoverId);
-        toolTipLines = params.getMultilineText("tooltip", toolTipLines);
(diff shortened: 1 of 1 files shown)


Nightenom:
This bug was fixed in blockui iirc, if true please apply same logic as in blockui

uecasm:
Can you point to the PR/commit where it was fixed in BlockUI?  The code related to this commit still appears broken in BlockUI, so if it was fixed it was via a completely different method.

uecasm:
Ok, I found it.  Weird I overlooked that when I previously checked the BlockUI code; maybe I had an older version or something.

PR #520, opened 2022-07-30
Command Tool

uecasm:
- Adds a recipe for the tag tool, since there's no particular reason it can't be used in survival.
    image
- Pressing ESC out of the scan tool GUI will now remember the coordinates and schematic name previously entered (until the scan is actually completed or canceled).
    - Previously, some random actions (editing the filter text or clicking the "show resources" button) would save the coords, but they would otherwise be reverted.  The name would never be saved, although there was a way to auto-populate it (once).
- The schematic name is also shown on the scan tool's tooltip:
    image
- Adds a new creative-only item, the Command Tool, intended to make it easier to set up and maintain auto-scanners.
    - Right-clicking it on a Command Block (with server op) will open the command block GUI as usual.
    - Shift-right-clicking it on a Command Block (or right-clicking without server op) will Copy.
        - This imprints the build's coords onto the command tool.
        - It also imprints the coords onto the closest scan tool on your hotbar.
        - It also shows the scan area in-world in the same manner as the scan tool.
    - Left-clicking it on a Command Block will Paste.
        image
        - This only requires creative mode, not server ops, to update the command.
        - The command block must either be empty or contain a scan command already.
        - If the existing scan command has a different building name, you just get a warning and it will not paste.  But you can override this by shift-left-clicking instead.
        - This fills in _your current scan tool region_ (coords and filename) as a scan, and also re-binds the command tool to this command block and build.  It need not be the same command block you copied from.
    - Right-clicking in air will teleport the player between the last-bound command block and just outside of the build's scan region.
        - This operates by internally executing a `/tp` command for the player.

Review please

So, basically, you can shift-right-click a command block to "load" the build coords, then right-click in air to teleport to the build, fiddle with it, and teleport back.  When making a new build (or updating the area of an existing build), you can set up the area and name with the scan tool (but not actually scan it), then teleport back to your (previous) command block, and then paste onto a new one.

Currently, there are no settings added for this; but I've considered whether it might be useful to add some for one or more of the following (perhaps for other servers):
- Which direction, and how far away, you're teleported when teleporting to the build.
    - The current values work well for the single-player downloadable map.  I don't know if the server is laid out similarly or not.
- If you must be inside a max distance from (one of) the command block or build before it allows you to teleport.
- If teleportation should be disabled entirely.
- If updating coords should be restricted to server ops only.

I was also debating a little whether confirm/discard on scan tool should actually clear the name or not (since it doesn't completely clear the coords).  Currently it does, both for consistency with the scan tool itself (and preserving some previous behaviour) and as a reminder that you probably want to enter a new name.  But there could be some benefit in retaining it instead...

Caveat: teleporting will sometimes (on servers) display a "moved too quickly" warning in the console, although no other apparent ill effects.  This seems like a Mojang bug since it's just using the built-in command to actually perform the teleport.


src/main/java/com/ldtteam/structurize/items/ItemCommandTool.java
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    /**
+     * Searches the player's hotbar (only) for a scan tool.
+     * If there's multiple, it prefers the closest item to the current slot, starting from the left side.
+     * @param player The player to search.
+     * @return The scan tool found, or an empty stack.
+     */
+    @NotNull
+    private ItemStack findScanTool(@NotNull final Player player)
```
Nightenom, line 276: There is similar logic in mcol for searching any item in inv, the double for is not needed iirc (same behaviour is provided by implementation of player inv iirc)

src/main/java/com/ldtteam/structurize/items/ItemCommandTool.java
```diff
@@ -0,0 +1,464 @@
...
+     * It's usually most convenient to appear outside the scanned area (it might be filled with placeholders),
+     * so we just pick an arbitrary location on a particular side that's convenient.  Perhaps this
+     * could be made into a setting.
+     *
+     * @param from The first bounding corner.
+     * @param to The second bounding corner.
+     * @return A convenient teleportation position near but not inside the scan area.
+     */
+    @NotNull
+    private static BlockPos getTeleportPos(@NotNull final BlockPos from,
```
Nightenom, line 309: Seems sus, can we just use the mechanism for finding spawn or sth like that (there are/were many things for finding possible spawn pos in vanilla)

src/main/java/com/ldtteam/structurize/network/messages/UpdateAutoScannerMessage.java
```diff
@@ -0,0 +1,101 @@
...
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Updates the command in a command block auto-scanner.
+ */
+public class UpdateAutoScannerMessage implements IMessage
```
Nightenom: Better name please, auto scanner is our specific usage, something like UpdateScanCommandBlock would fit better

src/main/java/com/ldtteam/structurize/network/messages/ShowScanMessage.java
```diff
@@ -0,0 +1,103 @@
...
+        this.anchor = anchor;
+        this.fileName = fileName;
+    }
+
+    /**
+     * Deserialize.
+     */
+    public ShowScanMessage(final FriendlyByteBuf buf)
+    {
+        final FriendlyByteBuf buffer = new FriendlyByteBuf(buf);
```
Nightenom: what?

src/main/java/com/ldtteam/structurize/items/AbstractItemWithPosSelector.java
```diff
@@ -138,4 +139,18 @@ public float getDestroySpeed(final ItemStack stack, final BlockState state)
...
         return Float.MAX_VALUE;
     }
+
+    /**
+     * Saves the start/end coordinates on this stack.
+     * @param tool The tool stack (assumed already been validated)
+     * @param start The new start position
+     * @param end The new end position
+     */
+    public static void setBounds(@NotNull final ItemStack tool,
```
Nightenom, line 149: i would make this non-static and do if(item instanceof scantool tool) tool.setbounds(stack, start, end)

src/main/java/com/ldtteam/structurize/items/ItemCommandTool.java
uecasm: Code in mcol is not helpful, since the dependency points the other way.

The double for is entirely necessary (as is not using other code) because this implements a very specific search order to find the "correct" scan tool, in case someone is carrying more than one.  This is documented.

src/main/java/com/ldtteam/structurize/items/ItemCommandTool.java
uecasm: I don't want a random position vaguely around the build area, I want a very specific point at at specific side of the build area (where the signs are located).  Also, I want something that can calculate a coordinate without requiring that the corresponding chunk be loaded at the time (so it can't inspect the world to find a hole).

src/main/java/com/ldtteam/structurize/network/messages/ShowScanMessage.java
uecasm: Fair point.  In my defense, I was following what this code does:

https://github.com/ldtteam/Structurize/blob/3d9952bb620656adee0c6a727e291722a5738cca/src/main/java/com/ldtteam/structurize/network/messages/SaveScanMessage.java#L30-L33

But you're correct that this is silly, and I'll change my one.

src/main/java/com/ldtteam/structurize/network/messages/UpdateAutoScannerMessage.java
uecasm: To me "auto scanner" == "scan command block", but ok.

src/main/java/com/ldtteam/structurize/items/AbstractItemWithPosSelector.java
uecasm: That would just make it much harder to use it.  Not helpful.

src/main/java/com/ldtteam/structurize/items/ItemCommandTool.java
Nightenom: 1) you can copy the code to here, we do that for many things
2) i understood the intention, but two things: first - you assume that player inventory has 27 slots - i definitely dont need to tell you that other mods may do virtually anything, and second - I would just perform same scan what middle mouse click does (aiming at "use already something what players normally use")

src/main/java/com/ldtteam/structurize/items/ItemCommandTool.java
Nightenom: Yeah, you can use this result as center of search area, but I don't see any check that this blockpos is useable as destination = at least two vertical air block hole

src/main/java/com/ldtteam/structurize/items/AbstractItemWithPosSelector.java
Nightenom: yeah, but overrideable

src/main/java/com/ldtteam/structurize/network/messages/ShowScanMessage.java
Nightenom: Super weird, might be some leftover from older times when duplicating was done for data security? w/e

src/main/java/com/ldtteam/structurize/items/AbstractItemWithPosSelector.java
uecasm: To elaborate, here is a usage case (in `ItemCommandTool`):

```java
        final ItemStack scanTool = findScanTool(player);
        if (scanTool.isEmpty()) { /* error out */ }
```

then many many lines later:

```java
            ItemScanTool.setBounds(scanTool, from, to);
            ItemScanTool.setAnchorPos(scanTool, anchor);
```

To do what you suggest would require opening a large nested block, or swapping `scanTool.isEmpty()` for the much more long-winded `!(scanTool.getItem() instanceof ItemScanTool scanToolItem)`, just to permit using `scanToolItem` further down.

Also, items are singletons anyway, so if non-static it could be called with `ModItems.scanTool.get().setBounds(...)`, but I don't think that's better than just calling the static method either.

Your suggestion does have the merit of somewhat validating that the stack probably contains a scan tool (which was I assume what you were going for), but in all places that these are called that's already been validated by other means.  Hence just stating that it's a precondition of the method in the javadoc.

----
Another usage site is more simple (in `UpdateScanToolMessage`):

```java
        if (stack.getItem() == ModItems.scanTool.get())
        {
            ItemScanTool.setBounds(stack, from, to);
        }
```

And there it could more easily do what you're suggesting, but I don't think it's better.  The one caveat is that your version potentially allows for the `ItemScanTool` to be subclassed and it would still find it; but I'm not sure why that's a thing anyone would want to do.  And indeed I think it's safer to *not* support that case until there's a good motivation for it.

Nightenom (dismissed):
based on discord convo:
Just try to check about the tp being at least a bit smarter = don't tp me into wall
Then mergeable

uecasm:
I'm open to suggestions of existing methods to check/choose an alternative location.  I don't want to do it at copy time since I don't want that to load the chunk, but it could do it at actual teleport time.  I've had a quick look around the vanilla code but haven't found anything that looks promising (but it's not hard to miss things).

I did test out one thing, but somewhat as expected, when teleporting to an unloaded chunk it causes Very Bad Things™ (it makes you teleport to y=-64 instead of the correct location, and then you die, and probably lose all your stuff).  So yeah, that's not going to fly.  The chunk doesn't get loaded until the teleport actually occurs.

Currently though it assumes that you're building things using an auto-scanner setup similar to the single-player world download (and presumably similar to the schematic server); as such, the two positions it uses are "standing on top of the command block" and "3 blocks to the south of the center of the building", which should always be safe places (unless the scan coords are totally wrong).

uecasm:
Ok, this seems to work not too terribly.

Nightenom (dismissed):
Please rebase and @Raycoms mergeable

uecasm:
Now updated for compatibility with the new build tool.  This made some things easier and others harder, but the functionality is the same, just a little different under the hood.

uecasm:
FYI I'm experimenting with a completely different approach, so don't merge either of these just yet.

PR #523, opened 2022-08-11
Massive hack to scan entities, at least it works

Raycoms:
- This is a massive hack to allow scanning entities that are in unloaded chunks.
- I tested this several times in different combinations.
- I believe, that as of the current system that's the best we can do.
- In the future we might want to think about alternatives that are easier to maintain like:

-- Not supporting scanning in unloaded chunks at all
-- Async scanning (chunk load the area, hand it to a ticker that waits properly for the entities to be loaded)
-- etc

### DIFF
--- a/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
+++ b/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
@@ -119,7 +119,7 @@ public class BlueprintUtil
                 {
                     continue;
                 }
-                entitiesTag.add(entity.serializeNBT().copy());
+                entitiesTag.add(entity.serializeNBT());
             }
 
             world.entityManager.permanentStorage.flush(false);
--- a/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
+++ b/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
@@ -119,7 +119,7 @@ public class BlueprintUtil
                 {
                     continue;
                 }
-                entitiesTag.add(entity.serializeNBT());
+                entitiesTag.add(entity.serializeNBT().copy());
             }
 
             world.entityManager.permanentStorage.flush(false);
@@ -147,7 +147,7 @@ public class BlueprintUtil
                                     final Vec3 posVec = new Vec3(posList.getDouble(0), posList.getDouble(1), posList.getDouble(2));
                                     if (area.contains(posVec))
                                     {
-                                        entitiesTag.add(entityCompound);
+                                        entitiesTag.add(entityCompound.copy());
                                     }
                                 }
                             }
--- a/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
+++ b/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
@@ -3,19 +3,20 @@ package com.ldtteam.structurize.blueprints.v1;
 import com.ldtteam.structurize.api.util.BlockPosUtil;
 import com.ldtteam.structurize.api.util.Log;
 import com.ldtteam.structurize.blockentities.interfaces.IBlueprintDataProviderBE;
+import net.minecraft.core.SectionPos;
+import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.block.Blocks;
 import net.minecraft.world.entity.Entity;
-import net.minecraft.world.entity.decoration.HangingEntity;
 import net.minecraft.nbt.*;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.SharedConstants;
 import net.minecraft.util.datafix.fixes.References;
 import net.minecraft.util.datafix.fixes.ChunkPalettedStorageFix;
+import net.minecraft.world.level.chunk.storage.EntityStorage;
 import net.minecraft.world.phys.AABB;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.phys.Vec3;
-import net.minecraft.world.level.Level;
 import net.minecraft.server.level.ServerLevel;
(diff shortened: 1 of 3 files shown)


src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
```diff
@@ -109,43 +110,82 @@ public static Blueprint createBlueprint(
...
 
             final ListTag posList = new ListTag();
             posList.add(DoubleTag.valueOf(oldPos.x - pos.getX()));
             posList.add(DoubleTag.valueOf(oldPos.y - pos.getY()));
             posList.add(DoubleTag.valueOf(oldPos.z - pos.getZ()));
 
-            BlockPos entityPos = entity.blockPosition();
-            if (entity instanceof HangingEntity)
+            entity.put("Pos", posList);
+            if (entity.contains("TileX"))
```
Raycoms, line 177: Btw, this new handing is already superior to the old handling in this regard.

src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
```diff
@@ -109,43 +110,82 @@ public static Blueprint createBlueprint(
...
+                            final CompoundTag compoundTag = ((EntityStorage) world.entityManager.permanentStorage).worker.load(new ChunkPos(x, z));
+                            if (compoundTag == null)
+                            {
+                                continue;
+                            }
+
+                            final ListTag tagList = compoundTag.getList("Entities", CompoundTag.TAG_COMPOUND);
+                            for (int i = 0; i < tagList.size(); i++)
+                            {
+                                final CompoundTag entityCompound = tagList.getCompound(i);
```
Nightenom, line 143: deep copy?

src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
```diff
@@ -109,43 +110,82 @@ public static Blueprint createBlueprint(
...
-        for (final Entity entity : entities)
-        {
-            if (!entity.getType().canSerialize())
+            world.entityManager.permanentStorage.flush(false);
+            for (int x = SectionPos.blockToSectionCoord(pos.getX()); x <= SectionPos.blockToSectionCoord(pos.getX() + sizeX); x++)
             {
-                continue;
+                for (int z = SectionPos.blockToSectionCoord(pos.getZ()); z <= SectionPos.blockToSectionCoord(pos.getZ() + sizeZ); z++)
+                {
+                    if (!world.entityManager.areEntitiesLoaded(new ChunkPos(x, z).toLong()))
```
Nightenom, line 130: Imho this is not time safe, in means between world.getEntities(null, area) and for { for { blah }} can the chunk turn from unloaded to loaded and data will be missed

Raycoms:
Within the same tick?

Sent from Proton Mail mobile

src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
Nightenom: > Within the same tick?

well, idk how async io is done there

Raycoms:
The io result will be picked up and processed in the next tick

Sent from Proton Mail mobile

src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
Nightenom: > The io result will be picked up and processed in the next tick

Then it's fine, just the nbt copy, since were changing it

marchermans:
So why was this added if we are considering just not supporting it? Additionally I don't think scanning of an unloaded area should be supported in the first place....

uecasm:
> Additionally I don't think scanning of an unloaded area should be supported in the first place....

Scanning of an unloaded area is somewhat inevitable with the command block scanning rooms being some distance away from the buildings they're scanning.  (Both in the schem server and in standalone scanning worlds.)  It would be less of an issue if the command block was next to the building being scanned, but then there'd be no way to do a "scan all" since some of the command blocks (and redstone) would be unloaded.

Unless perhaps the actual scanning command blocks are always next to the scans, and the "scan all" used a completely different set of command blocks that somehow triggered the chunk load and executed the command in the other scan block.  But that seems like a maintenance nightmare.

marchermans:
> > Additionally I don't think scanning of an unloaded area should be supported in the first place....
> 
> Scanning of an unloaded area is somewhat inevitable with the command block scanning rooms being some distance away from the buildings they're scanning.  (Both in the schem server and in standalone scanning worlds.)  It would be less of an issue if the command block was next to the building being scanned, but then there'd be no way to do a "scan all" since some of the command blocks (and redstone) would be unloaded.
> 
> Unless perhaps the actual scanning command blocks are always next to the scans, and the "scan all" used a completely different set of command blocks that somehow triggered the chunk load and executed the command in the other scan block.  But that seems like a maintenance nightmare.

I know these commands exists but mojang has argument types which force a position to be loaded to be valid. And mojang uses them for example in their fill and replace commands. In my opinion that is simply the guidelines we should follow instead of implementing this hack

src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
```diff
@@ -119,7 +119,7 @@ public static Blueprint createBlueprint(
                 {
                     continue;
                 }
-                entitiesTag.add(entity.serializeNBT());
+                entitiesTag.add(entity.serializeNBT().copy());
```
Nightenom: I would say u don't have to copy new instance? The second was the one I wanted

Nightenom (approved):

Raycoms:
Yeah but mojang doesn't scan in hundreds of blueprints

Raycoms:
Because that's what it takes to scan a style on the schematic server

Sent from Proton Mail mobile

PR #525, opened 2022-08-22
better sorting

Raycoms:
Closes #
Closes #
Closes #

- Random sort pack list once per session
- Also order categories
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
+++ b/src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
@@ -14,7 +14,9 @@ import net.minecraft.ChatFormatting;
 import net.minecraft.network.chat.TextComponent;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Random;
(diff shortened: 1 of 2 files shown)


src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
```diff
@@ -42,6 +44,11 @@ public class WindowSwitchPack extends AbstractWindowSkeleton
      */
     private List<StructurePackMeta> packMetas;
 
+    /**
+     * Random with a fixed random seed.
+     */
+    private static int randomSeed = new Random().nextInt();
```
marchermans, line 50: Just set it to 42 instead of doing this.

marchermans (changes requested):

src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
marchermans: Disregard.

marchermans (approved):

PR #528, opened 2022-08-26
Fix placement of settings menu in shape tool

uecasm:
- Fixes settings menu appearing incorrectly in shape tool gui
img

Review please

~~Still not really sure why this was only happening in the shape tool and not the build tool, since they share code and xml.  But this doesn't break the build tool, and it seems like a more reasonable number anyway, so...~~


Raycoms:
I'd handle this like I did in the supplycamp window, move it on opening the shape tool. In the buildtool the current location is good.

uecasm:
This moves it to appear in the same spot as it appears in the build tool.  Though like I said, I couldn't find any code that explains why it worked in the build tool before.

uecasm:
The settings button itself is explicitly positioned to a y-coord closer to 150 than to 50 anyway, so it makes more sense this way.

Raycoms:
ah ok

uecasm:
Oh wait, now I did find the code.  The build tool moves it to (5, 140), so it's close but not quite identical to this.  I'll make them the same.

PR #534, opened 2022-09-17
add paste command

Raycoms:
Closes #
Closes #
Closes #

- Add a command to paste from
-
-

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/commands/PasteCommand.java
+++ b/src/main/java/com/ldtteam/structurize/commands/PasteCommand.java
@@ -1,0 +1,225 @@
+package com.ldtteam.structurize.commands;
+
+import com.ldtteam.structurize.blocks.interfaces.ISpecialCreativeHandlerAnchorBlock;
+import com.ldtteam.structurize.blueprints.v1.Blueprint;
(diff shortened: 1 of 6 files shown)


someaddons (approved):

src/main/java/com/ldtteam/structurize/commands/PasteCommand.java
```diff
@@ -0,0 +1,225 @@
...
+
+        return execute(context.getSource(), pos, packName, path, rotation, mirror, pretty);
+    }
+
+    protected static LiteralArgumentBuilder<CommandSourceStack> build()
+    {
+        return newLiteral(commandName)
+          .then(newArgument(POS, BlockPosArgument.blockPos())
+            .then(newArgument(PACK_NAME, StringArgumentType.word())
+              .then(newArgument(FILE_PATH, StringArgumentType.word())
```
uecasm, line 215: These should use the `string` argument type, so you can write a quoted string with spaces instead of doing the dot-replacement thing.  That's just weird.

Raycoms: didn't know that was a thing

PR #535, opened 2022-09-17
Multi-Scan, Command, and Teleport

uecasm:
Closes #520
Closes #533

- Adds a recipe for the tag tool, since there's no particular reason it can't be used in survival.
    image
- Pressing ESC out of the scan tool GUI will now remember the coordinates and schematic name previously entered.
    - Previously, some random actions (editing the filter text or clicking the "show resources" button) would save the coords, but they would otherwise be reverted.  The name would never be saved, although there was a way to auto-populate it (once).
- The schematic name is also shown on the scan tool's tooltip:
    image
    (actually this pic is a bit outdated, it shows a slot number on there now too)
- Enhanced the scan tool with multi-scan support:
    - A single scan tool now contains 10 slots, numbered 1-9 plus 0.  (If you really want more slots, you can use more than one tool; they're all independent.)
    - Each slot holds one scan area (two corners), optional anchor, and a name.
    - Anything can be stored in any slot; you can have the same name and/or coords in multiple slots if you like.
        - Suggested usage is to put the "real" scan boxes for level 1-5 in slots 1-5, and use slots 6-0 for replace boxes and other temporary purposes.  But you can do whatever works best for you.
    - The coords are (almost) always saved; certain changes are reverted if you click the red X but otherwise (even if you click the X) it remembers the coords and the name in each slot.
        - Caveat: the auto-name-from-anchor will only work if you clear the name manually first.  But I suspect many people didn't even know this was a thing anyway.
    - There are three different ways to switch between the slots:
        - You can hold shift and mouse-wheel up or down with the scan tool selected.
        - You can pick-block (def. middle-click) in air to switch up or sneak-pick-block in air to switch down.
        - With the scan tool GUI open, you can press the number keys 0-9 to switch to that slot.
            - Note that if you have input focus in a field (e.g. the scan name), the keys will type into that instead.
            - To allow you to use the hotkeys again, the first press of ESC will clear focus, and the second will close the window.
    - The current slot number and schematic name are shown on the item tooltip as well as on the hotbar tip, which reappears whenever you switch slots.  Switching slots will also highlight the scan area in-world.
        multiscan
- Implemented command block copy/paste (creative only):
    - Using pick-block with the scan tool on a command block will copy the command's scan into the tool's current slot.  (Any other block will actually pick-block, not switch slots.)
    - Using sneak-pick-block with the scan tool on a command block will paste the current slot into the command block.
        - This only requires creative mode, not server op, assuming that the command block itself is already placed.
        - As a safety feature, it will only overwrite a scan command if the name is the same, unless you hold CTRL as well to force it.
- Implemented scan teleportation (creative only):
    - Added a reconfigurable keybinding (unset by default) to trigger teleport.
    - If you press the keybind with a scan tool selected in your hotbar, it will teleport you to whichever is the furthest away of the last-bound command block and just outside the area of the current scan slot.  (Essentially toggling you between the two.)
        - I considered, but did not implement, having separate keybinds such that one will always teleport you to the command blocks and one would always teleport you to the current build.  In particular, the latter would permit teleportation even if you never bind any command blocks, which is not currently supported.
    - A new command block position is bound whenever you copy or paste a command block.  It only remembers the most recent command block across all slots, not one per slot.  (The assumption is that the command blocks will be reasonably close to each other, or at least that you will be working mostly with one related line of blocks that are close to each other.)
    - There are server configuration options to adjust the teleportation location, or to disable teleportation entirely.
    - I considered, but did not implement, an option to restrict how far away you can teleport.  (I figure, it's creative anyway.)
- Allows slashes to be included in the scan command block filename (matching the scan tool GUI), so that subdirectories can be included in the scan too.
- A few misc tidy-ups of related code.

Review please

This is an alternative implementation, based on Moobien's suggestion to support multiple concurrent scans and Nightenom's to just enhance the scan tool instead of making a new command tool.  (It seemed really silly to have multi-scans **not** in the scan tool itself.)

It's currently in 1.19 only, but I'll make a backport if everyone seems happy with this as is.  Since this was a rewrite I broke things up into smaller commits as well; hope that's helpful. 😁


src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
```diff
@@ -151,15 +166,87 @@ public static void onClientTickEvent(final ClientTickEvent event)
...
+            }
+        }
+
+        mc.getProfiler().pop();
+    }
+
+    @SubscribeEvent
+    public static void onMouseButtonClick(final InputEvent.MouseButton.Pre event)
+    {
+        // it would have been nice to use the "pick block" event instead, but that only fires if clicked on a block
```
Nightenom: well, since u speaking about u might use pick block keybinding value instead of glfw_mouse_button_3

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
```diff
@@ -241,12 +233,479 @@ public boolean canAttackBlock(final BlockState state, final Level worldIn, final
...
+        if (stack.hasTag())
+        {
+            tooltip.add(getCurrentSlotDescription(stack));
+        }
+    }
+
+    @Override
+    public Component getHighlightTip(@NotNull final ItemStack stack, @NotNull final Component displayName)
+    {
+        return Component.empty()
```
Nightenom, line 254: in previously versions using component.empty() equivalent for creating new components wasn't safe, does this method creates new instance of component or returns static singleton?

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
```diff
@@ -241,12 +233,479 @@ public boolean canAttackBlock(final BlockState state, final Level worldIn, final
...
+        {
+            // it's always ok to set a blank command block
+        }
+        else if (!command.getCommandBlock().getCommand().contains(MOD_ID + " " + ScanCommand.NAME + " "))
+        {
+            player.displayClientMessage(Component.translatable("com.ldtteam.structurize.gui.scantool.paste.badcommand"), false);
+            player.playNotifySound(SoundEvents.NOTE_BLOCK_BIT, SoundSource.PLAYERS, 1.0F, 1.0F);
+            return;
+        }
+        else if ((modifiers & GLFW.GLFW_MOD_CONTROL) == 0)
```
Nightenom: wut, screen.isCtrl/shift/altDown() please
(or since this seems to server side then copy out it's logic with comment for future reference)

Nightenom (dismissed):
Too exhausted to do full reviews (leaving it for others), so just few notes
happy u created find safe tp method

src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
uecasm: I considered that, but I worried it might be set to a non-mouse-button, and then things get a lot more complicated.

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
uecasm: The `ComponentContents` is a singleton, the outer `MutableComponent` is not.  (And appending modifies the latter, not the former.)  So I'm reasonably sure this is safe.  Certainly appears to work fine.  Definitely something to bear in mind when backporting, though, that code is different in 1.18 so it might be unsafe there.

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
uecasm: It passes the `modifiers` across from the `event.getModifiers()`, so this seems more correct as is.  This isn't a GUI, there is no screen.

src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
Nightenom: Well, do that in client tick then :D generally speaking I thought that input events are for modifying/stopping inputs before any logic, thus u should not? use them for logic

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
uecasm: I was originally hoping to use `player.isSprinting()` but sadly that only triggers when actually sprinting, not merely when holding the sprint key.

src/main/java/com/ldtteam/structurize/event/ClientEventSubscriber.java
uecasm: The logic it does is to decide whether to let it continue to do the regular pick block (or other middle mouse bound action) or not, so it's mostly necessary there.  The actual work is passed off to the server via packet though.

uecasm: I've changed it now to use the block pick keybind, and `Screen.hasControlDown()`.

src/main/java/com/ldtteam/structurize/items/ItemScanTool.java
uecasm: I've found some vanilla code that uses the exact same pattern to create new mutable components, so it's definitely safe.  Will still keep it in mind for the backport though.

uecasm:
_coughs_

Raycoms (approved):

PR #538, opened 2022-10-02
Keybinds

uecasm:
- Changes blueprint preview keys to reconfigurable keybinds
    - The defaults are the same as previously, except that - and + are mapped specifically to the numpad keys now instead of supporting either.  But you can rebind them if you don't have a numpad.

My preferred bindings:
image

Review please

~Draft pending actual BlockUI version update from ldtteam/BlockUI#28.~


Nightenom:
Well, honestly since keybinds causes logic then it should sit in GUI tick, which should also allow u to use them without removing the blockui filter, which exists because of key and char handlers being connected together, removing it should causes sth like doubled typing or doubled ctrl+v actions (I don't exactly remember why I've introduced it and I also thought about removing it yr ago, but I didn't manage to test it properly)

uecasm:
According to the [official docs](https://docs.minecraftforge.net/en/1.19.x/misc/keymappings/#inside-a-gui), client tick is only for in-world keybinds.  GUI keybinds are supposed to be handled from the key press handler, which is what this does (once the filter is removed).

I have tested this and not found any doubled input that I can see, it just works.  (Although to clarify: I've only tested it with Structurize alone thus far, not with MineColonies.)

uecasm:
FWIW, the flow of events is that the key press handler occurs first, which in turn calls `window.onKeyTyped(0, key)`.  If handled at this point, it returns true, and nothing further happens -- this will be the case for keybinds and manual special key checks.

If that returned false (didn't handle it), the character handler occurs next, which in turn calls `window.onKeyTyped(ch, key)`.  (Except for whatever reason at this point, `key == 0` always as far as I can tell.)  It might repeat some of the same logic but that would all pass anyway, and only do logic related to the new value of `ch`.

Raycoms (dismissed):

uecasm:
Updated for new BlockUI build (note as of this writing it's not visible on CF yet...)

Nightenom (approved):

PR #540, opened 2022-10-08
Tag anchor substitution

uecasm:
- render placeholders "OFF" now hides placeholders and "ON" shows them (setting was previously reversed)
    - solid placeholders still always render as themselves, since the existing code to render them as their target block was explicitly disabled as it only worked server-side (or in single player), and I didn't want to change that
- enable datagen, with the intent to move more stuff there later
- fixes broken responses when using the tag tool (I don't know what `LanguageHandler` is trying to do, but it does parameters very badly)
    - actually, since there weren't many usages left, just removed use of it for translation entirely (all the newer code was avoiding it anyway)
- fixes the output messages of the calipers -- it's not legal to build a single sentence from individual word translations
- supports the "invisible" tag on all schematics (previously this required implementing a special interface, and currently mcol only does this for buildings and not for decorations anyway; that handling can be removed after bumping)
- changes the "tag anchor" block to support a configurable replacement block that it will turn into when built/pasted, instead of always turning into air
    - (this was needed primarily to allow the mcol mine blueprints to be tagged as invisible without changing their current anchor positions, but it may be useful in other cases as well)
    - the default state is now transparent, to indicate that it will replace with air
    - to "absorb" a replacement block, pick-block the tag anchor item on the desired block in the world
        - all plain blocks are supported, and a limited whitelist of block entities (currently only chests, signs, lecterns, and DO blocks)
    - the replacement block is shown in the hotbar selection and item rendering
      image
      (with some limited caveats: e.g. chests and signs don't render in hand, due to the way they're implemented)
    - the replacement block is shown in the item tooltip
      image

anchorview

Review please

Note this is the first time I've attempted custom rendering, so I might have done something more complicated than needed.  But it seems to work.

Also note that currently the rotation/mirror support is limited to doing so within a blueprint.  It will completely ignore you if you attempt to rotate it in-world using a wrench item or similar.  Similarly, your facing direction when placing the anchor block is also ignored.  Instead, rotate the original block to the desired facing before you absorb it.


marchermans (dismissed):
In principle this is Okey.
You do not need the custom model but this works (see chests for custom renderers with out one)
The other option would have been to dynamically bake the model with the replacement block Ala DO.

But if this is your first time: Nice work

src/main/java/com/ldtteam/structurize/blockentities/BlockEntityTagSubstitution.java
```diff
@@ -157,4 +178,190 @@ public void onDataPacket(final Connection net, final ClientboundBlockEntityDataP
...
+
+        /**
+         * Creates and loads (once) the replacement block entity, or returns the preloaded one.
+         * @param pos the blockpos to use (ignored if already loaded)
+         * @return the new or cached entity, or null if there isn't one
+         */
+        @Nullable
+        public BlockEntity getBlockEntity(final BlockPos pos)
+        {
+            this.cachedBlockentity = Optional.ofNullable(this.cachedBlockentity)
```
Nightenom: Honestly if(null) is more readable

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -132,21 +131,45 @@ private void init(final BlockPos anchorPos)
...
-                {
-                    state = BlockUtils.getWorldgenBlock(serverLevel, anchorPos.offset(blockPos), blueprint.getRawBlockStateFunction().compose(b -> b.subtract(anchorPos)));
-                    if (state == null)
+                    if (state.getBlock() == ModBlocks.blockSubstitution.get())
+                    {
+                        state = Blocks.AIR.defaultBlockState();
+                    }
+                    if (state.getBlock() == ModBlocks.blockTagSubstitution.get())
+                    {
+                        final Optional<BlockEntity> tagTE = tileEntities.stream()
```
Nightenom, line 142: iirc there is pos to te map in blueprint (tho it's probably compound) if not then create it, searching te list in at worst quadratic time (for each te you search te list again) has no place in rendering

Nightenom (changes requested):
avoid using java 8 stream things in rendering, otherwise same like last time: just quick code review, actual changes left to others

src/main/java/com/ldtteam/structurize/blocks/ModBlocks.java
```diff
@@ -65,6 +68,6 @@ public static <B extends Block> RegistryObject<B> register(String name, Supplier
         blockSubstitution       = register("blockSubstitution", BlockSubstitution::new, ModItemGroups.STRUCTURIZE);
         blockSolidSubstitution  = register("blockSolidSubstitution", BlockSolidSubstitution::new, ModItemGroups.STRUCTURIZE);
         blockFluidSubstitution  = register("blockFluidSubstitution", BlockFluidSubstitution::new, ModItemGroups.STRUCTURIZE);
-        blockTagSubstitution    = register("blockTagSubstitution", BlockTagSubstitution::new, ModItemGroups.STRUCTURIZE);
+        blockTagSubstitution    = BLOCKS.register("blockTagSubstitution".toLowerCase(), BlockTagSubstitution::new);
```
Raycoms, line 71: Why does this block go through a different path?

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
marchermans: Searching a list is at worst a linear time operation.
Sorting it is quadratic which is not what he is doing.

That said the stream is okey here Java 17 has intrinsics on this, but nights idea to use the map is better

Raycoms: Yeah, we def don't want to search this in each render tick

src/main/java/com/ldtteam/structurize/network/messages/AbsorbBlockMessage.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.core.BlockPos;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.fml.LogicalSide;
+import net.minecraftforge.network.NetworkEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class AbsorbBlockMessage implements IMessage
```
Raycoms, line 16: javadoc

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -289,18 +290,26 @@ public BlockPlacementResult handleBlockPlacement(
...
+                localState = tagEntity.getReplacement().getBlockState();
+                tileEntityData = tagEntity.getReplacement().getBlockEntityTag();
+            }
+            else
+            {
+                localState = Blocks.AIR.defaultBlockState();
+            }
+        }
+
+        if (BlockUtils.areBlockStatesEqual(localState, worldState, handler::replaceWithSolidBlock, handler.fancyPlacement(), handler::shouldBlocksBeConsideredEqual, tileEntityData, worldEntity))
```
Raycoms, line 310: why move this down?

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
```diff
@@ -32,86 +29,6 @@ private LanguageHandler()
...
-                        break;
-                    }
-                }
-
-                if (!isInArgs)
-                {
-                    translation.append(" " + object);
-                }
-            }
-        }
```
Raycoms, line 110: Will this cause an API issue with minecolonies?

Raycoms (changes requested):

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
Nightenom: its in init but still meh, and it gets quadratic since linear search multiplied by (at worst) all TEs in blueprint

uecasm: Searching is not sufficient, it has to remove the old BE from the list and substitute a different one.  I could probably change it to use a map instead of a flat list but that would touch a lot more of the existing code, and maps are slower than lists (for the _other_ operations).  The list of BEs is _usually_ fairly short (although using lots of DO blocks distorts this) so it seemed it would be faster as is, although admittedly I haven't actually measured.

Given that this code is literally reinstantiating all the block entities, a list search is going to be the least of the time consumed, and it's not happening every tick; this is all in a conditional cache.

And this search should happen at most once per blueprint.  There should normally only be one tag anchor block in each blueprint -- while you _can_ use more than one, there's not any benefit in doing so, so you wouldn't.

src/main/java/com/ldtteam/structurize/blocks/ModBlocks.java
uecasm: Because it's doing something different.  It's not making a plain `BlockItem`, it's making a subclass.
Or more correctly, it's avoiding registering the item here, because it's registering that separately over in `ModItems`.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
uecasm: No, I checked (by reading the code, not as yet running with it), which is why I left the `translateKey*` methods, as those are all that mcol uses.  See the javadoc change at the top of the file 😄

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
uecasm: Because it should calculate the target block of the placeholders before it decides whether or not the target block is already in the world.  Not doing that is why "placeholders" often show up in the resource requirements lists unnecessarily.

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
uecasm: MineColonies is mostly using `MessageBuilder` now instead, it's only the config-file stuff that's still using `LanguageHandler`, and the comments on that indicate it doesn't really want to be.

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
Nightenom: _Usually_ has no place in algorithms "sold" to anyone else
Tag anchor != Tag substitution block, or if that's true then tag substitution block is weirdly named class

uecasm: The in-code name follows the naming convention for the other placeholders.  Its translated name is more descriptive of its intended usage (also like the other placeholders).

src/main/java/com/ldtteam/structurize/util/LanguageHandler.java
Raycoms: kk

PR #551, opened 2022-10-27
Fix code warnings

uecasm:
- Fixes various static analyzer warnings
    - Closes directory enumeration streams properly
    - Avoids busy-waiting for structure pack loading
    - Avoids deadlocking if thread is interrupted during loading (e.g. client disconnect)

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
+++ b/src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
@@ -105,9 +105,10 @@ public class WindowSwitchPack extends AbstractWindowSkeleton
     @Override
     public void onOpened()
     {
-        while (!StructurePacks.hasFinishedLoading())
+        if (!StructurePacks.waitUntilFinishedLoading())
         {
-            // Wait until finished loading!
+            close();
+            return;
         }
 
         // Here we would query from the online schematic server additional styles then, which, on select, we'd download to the server side.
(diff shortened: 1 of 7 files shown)


Raycoms (approved):
Is the barrier really necessary? Isn't an atomic/volatile boolean enough?

uecasm:
The barrier class as a whole is needed so that it can formally wait instead of busy-waiting (sleep loop).

The `synchronized` blocks are perhaps overkill given that we never actually call `close()` in this particular case, but the general pattern is recommended to introduce the correct write barriers, which a volatile variable itself does not.  Using an atomic (but not volatile) boolean might be a sufficient alternative, but since this is only ever used on the `IOPool` I don't think it's a worry.

Raycoms:
> The barrier class as a whole is needed so that it can formally wait instead of busy-waiting (sleep loop).
> 
> The `synchronized` blocks are perhaps overkill given that we never actually call `close()` in this particular case, but the general pattern is recommended to introduce the correct write barriers, which a volatile variable itself does not. Using an atomic (but not volatile) boolean might be a sufficient alternative, but since this is only ever used on the `IOPool` I don't think it's a worry.

I'd make it an atomic boolean then, it's only ever written once and I'm not a fan of locks =D

uecasm:
> I'd make it an atomic boolean then, it's only ever written once and I'm not a fan of locks =D

Well, `wait` requires the lock to be held, so blame Java.  Could possibly skip it with a fast-path double-checked lock, but that's technically an antipattern (though again probably harmless in this specific case).  Being anti-lock is entirely reasonable in game code, but that doesn't really apply on a background thread which already does a lot of blocking.

uecasm:
_coughs_

marchermans (approved):

uecasm:
_bumps_

PR #552, opened 2022-10-27
Local blueprints and style filters

uecasm:
Depends on #551 (so will need to be rebased after that is merged, hence being a draft; or I guess this could be merged into that immediately prior, if that's easier and everyone's happy with this)

- If a stylepack folder contains both blueprints and subfolders, currently the blueprints are hidden.  This instead shows a "." subfolder in that case, which if selected will show just the blueprints.  (This is most useful for the player's scans, but can happen accidentally in another stylepack.)
    - e.g. if `infrastructure/foo.blueprint` and `infrastructure/roads/bar.blueprint` exist, then you will see the following tree:
        - Infrastructure
            - .
                - foo
            - Roads
                - bar
- Allows (but does not require, for backwards compatibility) the style pack selection window to accept a predicate to hide some of the style packs.
    - Test case for this (not yet pushed to mcol) was to make the Supplies window hide stylepacks that don't have the corresponding blueprint -- e.g. if a pack only has a camp and not a ship, then it will appear when using a camp and not when using a ship.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
+++ b/src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Random;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import static com.ldtteam.structurize.api.util.constant.Constants.MOD_ID;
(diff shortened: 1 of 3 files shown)


src/main/java/com/ldtteam/structurize/client/gui/WindowSwitchPack.java
```diff
@@ -37,6 +38,11 @@ public class WindowSwitchPack extends AbstractWindowSkeleton
      */
     private final Supplier<BOWindow> prevWindow;
 
+    /**
```
Raycoms, line 41: Not sure I understand the reason for this.

uecasm:
_also bumps_

Raycoms (approved):

PR #578, opened 2023-03-19
Allow previewing disabled buildings

uecasm:
Closes ldtteam/minecolonies-features#736

- When a building requirement is not met (no colony, missing hut item, missing research), instead of disabling the building button entirely, this keeps the disabled visual state but allows clicking on the button to display the preview anyway.
- When a disabled blueprint is selected in this way, the green tick button is hidden so that it's still not possible to actually build the building.
    - This is not quite the same behaviour as in 1.18 or earlier (there, the hut block can actually be placed but the builder will refuse to start construction), so it is perhaps still not quite as good for colony planning purposes, but this will at least allow previewing all buildings without creative mode.
    - Restoring 1.18 behaviour is possible but would likely require a change to the interface that will mean changes to mcol will be needed as well, and I was trying to keep this self-contained.

Review please (do not backport, can port forward)


Raycoms (dismissed):

Talyda:
I love you I love you I love you!

No denying, 1.18 behaviour would be better, but this is SO much better than current!  Thank you!

Raycoms (approved):

Nightenom (approved):

PR #579, opened 2023-03-24
Hide the creative-only DO blocks from the replace window

uecasm:
- Removes the creative-only (texture-shifting) DO blocks from the scan tool replace window, to reduce likelihood of accidentally including them in schematics.
    - "Real" DO blocks that you have in your inventory still appear in the list.

Review please


uecasm:
Side note: while testing this I noticed that scrolling quickly to the end of the block list and clicking the Select button will usually select entirely the wrong block.  Clicking again after waiting a bit resolves this.  I haven't looked too closely but I assume this is a BlockUI issue.

marchermans (approved):

PR #581, opened 2023-05-13
Remove simple type equality and calculate resource delta

### LINKED ISSUE #580: Copper framed blocks replaced for free

- [X] I have searched the existing issues.

- [X] I am running the latest **alpha** version of Structurize for my Minecraft version.

- [X] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/).

### Minecraft Version

1.19

### Structurize Version

1.19.2-10.491-alpha

1.19.2-10.405

1. 43.2.0 Forge
2. no idea what you mean by block UI mod, JEI's version is 11.6.0.1105
3. 1.19-1.0.83 ALPHA

### Current Behavior

When building in the steampunk style, builders will replace any framed copper blocks with exposed (or more eroded copper types) framed blocks for entirely free, as long as the block was previously placed down before

they will not request these blocks

i can confirm this because i did not make exposed copper framed blocks, yet they were placed at my builders hut

### Expected Behavior

the builder would request exposed copper framed blocks, and use them to replace the old blocks

### Reproduction Steps

Make any building that uses framed copper blocks
Supply it with the needed materials (dont use creative quick build)
let the builder make it normally
repeat this until the building is at a level it requires framed exposed copper blocks
upgrade it, but do not give the builder any exposed copper framed blocks
wait and watch

### Logs

N/A

### Anything else?

example of what i mean by framed copper and framed exposed blocks

### Footer

---
## Viewers

uecasm:
Closes #580

- Removes simple block type equality check and early exit for resource requirements.
- Instead calculates a proper delta between the resources for the existing block vs. the resources for the new block (when they're the same type).
- Removes a stair placement handler whose only purpose in life appears to be to ignore blockstate changes to stairs for no readily apparent reason.  (Though oddly not ignoring changes to only one of the three properties of stairs.)
    - This appears to be the reason why stairs weren't flipped upside down during upgrade/paste unless they also changed block type.

Review please

Posting as a draft since this probably needs more testing, though I've tested it with MineColonies and with likely problematic blocks like racks and chests in addition to DO blocks.  Suggestions welcome for additional cases, or you can try it out yourself from the PR build.

Given a blueprint that has DO blocks that vary only by materials, this will now consider these to be "different" and will show them as requirements in the resource list in both the upgrade preview and the builder's request lists.

Most blocks that are the same blockstate but different NBT will be left untouched via [`BlockUtil.areBlockStatesEqual`](https://github.com/ldtteam/Structurize/blob/1871e0c222113d0467be335ad6afad5dc9c6b451/src/main/java/com/ldtteam/structurize/util/BlockUtils.java#L373) -- which deliberately has an exception for DO blocks, which made me assume that the "free upgrade" behaviour was indeed unintended.

Most blocks that have different blockstates will be replaced, unless a placement handler says otherwise.  This is the potentially most risky part of the patch.  Notably, racks have such a handler.

One interesting consequence of these changes is that if the blueprint contains a chest or rack with contents, and the chest/rack already exists with different contents, then the resource requirements will calculate the blueprint contents not present in the world (but not the reverse) and will make the builder ask for them, but then will not actually put them in the chest (because it's too risky to update the contents).  But this doesn't seem entirely like a bad thing, and in most cases the blueprints have empty chests/racks this won't behave any different.

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
+++ b/src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
@@ -205,7 +205,6 @@ public class StructurePlacer
       CompoundTag tileEntityData)
     {
         final BlockState worldState = world.getBlockState(worldPos);
-        boolean sameBlockInWorld = worldState.getBlock() == localState.getBlock();
 
         if (!(worldState.getBlock() instanceof AirBlock))
         {
@@ -323,15 +322,9 @@ public class StructurePlacer
             {
                 final List<ItemStack> requiredItems = new ArrayList<>();
 
-                if (!sameBlockInWorld && !this.handler.isCreative())
+                if (!this.handler.isCreative())
                 {
-                    for (final ItemStack stack : placementHandler.getRequiredItems(world, worldPos, localState, tileEntityData, false))
-                    {
-                        if (!stack.isEmpty() && !this.handler.isStackFree(stack))
-                        {
-                            requiredItems.add(stack);
-                        }
-                    }
+                    requiredItems.addAll(placementHandler.getRequiredItemsVsWorld(world, worldPos, localState, tileEntityData, false));
 
                     if (!this.handler.hasRequiredItems(requiredItems))
                     {
@@ -341,7 +334,7 @@ public class StructurePlacer
 
                 if (!(worldState.getBlock() instanceof AirBlock))
                 {
-                    if (!sameBlockInWorld
+                    if (localState.getBlock() != worldState.getBlock()
                           && worldState.getMaterial() != Material.AIR
                           && !(worldState.getBlock() instanceof DoublePlantBlock && worldState.getValue(DoublePlantBlock.HALF).equals(DoubleBlockHalf.UPPER)))
                     {
@@ -365,7 +358,7 @@ public class StructurePlacer
                     return new BlockPlacementResult(worldPos, BlockPlacementResult.Result.SUCCESS);
                 }
 
-                if (!this.handler.isCreative() && !sameBlockInWorld)
+                if (!this.handler.isCreative())
                 {
                     for (final ItemStack tempStack : requiredItems)
                     {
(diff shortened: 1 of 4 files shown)


src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -323,15 +322,9 @@ public BlockPlacementResult handleBlockPlacement(
             {
                 final List<ItemStack> requiredItems = new ArrayList<>();
 
-                if (!sameBlockInWorld && !this.handler.isCreative())
+                if (!this.handler.isCreative())
                 {
-                    for (final ItemStack stack : placementHandler.getRequiredItems(world, worldPos, localState, tileEntityData, false))
-                    {
-                        if (!stack.isEmpty() && !this.handler.isStackFree(stack))
```
Raycoms, line 330: no isStackFree check anymore?

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -205,7 +205,6 @@ public BlockPlacementResult handleBlockPlacement(
       CompoundTag tileEntityData)
     {
         final BlockState worldState = world.getBlockState(worldPos);
-        boolean sameBlockInWorld = worldState.getBlock() == localState.getBlock();
```
Raycoms, line 208: the origin of the world equals check was to speed up upgrades where blockstates changed because of some neighbors. Like a fence that now touches something outside of the schematic can be ignored. Maybe we should only bypass this for blockentities then.

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -649,42 +648,6 @@ public List<ItemStack> getRequiredItems(
         }
     }
 
-    public static class StairBlockPlacementHandler implements IPlacementHandler
```
Raycoms, line 652: I remember some infini stair loops. We need to test this all carefully.

Raycoms (commented):
Needs a lot of testing to make sure this doesn't make building heaps slower or cause loops.

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
uecasm: Good catch.  I had that in the other usage but forgot here.

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
uecasm: Yeah, fences connecting outside of schematic is tricky.  Though perhaps we just need a "fences don't care about blockstates" exception?

Raycoms: All connecting blocks like fences, walls, panes, etc

PR #588, opened 2023-07-10
Add timeout to selection tools

someaddons:
- Adds a timeout to position selection, if taking longer than 120 seconds to select a position, the previous data is cleaned up
Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/items/AbstractItemWithPosSelector.java
+++ b/src/main/java/com/ldtteam/structurize/items/AbstractItemWithPosSelector.java
@@ -30,6 +30,12 @@ public abstract class AbstractItemWithPosSelector extends Item
     private static final String START_POS_TKEY   = "item.possetter.firstpos";
     private static final String END_POS_TKEY     = "item.possetter.secondpos";
     private static final String MISSING_POS_TKEY = "item.possetter.missingpos";
+    private static final String TIMESELECTED = "item.possetter.timeselected";
+
+    /**
+     * Default timeout of 120 seconds
+     */
+    private static final int TIMEOUT_DELAY = 20 * 60 * 2;
 
     /**
      * MC redirect.
@@ -104,6 +110,7 @@ public abstract class AbstractItemWithPosSelector extends Item
     @Override
     public InteractionResult useOn(final UseOnContext context)
     {
+        checkTimeout(context.getItemInHand(), context.getLevel());
         final BlockPos pos = context.getClickedPos();
         if (context.getLevel().isClientSide())
         {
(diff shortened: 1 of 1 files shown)


Nightenom:
Rationale?
And ideally disable in dev please

Raycoms:
Case you set pós 0 0 0 to a few thousand by accident. Would even make it 30s or config and even less

Nightenom:
From private convo with Ray: 
> If you accidentally select far away positions it has to gather a shitton of data and you can't move in this time

This can be easily solved by distance check along with volume check we have already, without interrupting user interface, not to mention that there is no msg telling player you removed data

someaddons:
If you're setting a new position a long while after the old, the olds data is most likely invalid and can cause issues like very huge selections, so it then resets the second position aswell

uecasm:
Yeah, I don't really understand the goal of this.  As far as I can see, it only really affects the scan tool (the tag tool also uses that base but overrides the methods to do its own thing) and I really don't want the scan tool ever forgetting anything that I've set no matter how long I've spent doing other things.  Nor do I want it to care that during this session I only ever left clicked one corner, because the other corner might still be "correct".

Raycoms:
Now I fully got the timeout =D. Anyway, I would probably make this a volume check with a max volume in the config.

uecasm:
Speaking of inconvenient data loss and scanning silly areas for resources: when I shift-left-click an anchor it overwrites the corners with the anchor's corners.  These are usually wrong, unless the anchor got there by pasting.  So when I habitually click two corners then the anchor I have to remember to re-click the two corners before opening the GUI or it selects something silly (and then tries to scan that range for resources).

This was probably more convenient before we had command-block copy/paste, but now I find it only annoying. 😢

MotionlessTrain:
To add to this: My mouse sometimes registers my middle click as a left or right click (when I'm just not clicking exactly in the middle of my mouse). I have had times where I tried to paste in a schematic in a command block, and shift right clicked it instead, setting the second position by accident. Which is already annoying enough as I need to teleport back, reselect the placeholder block, and go back to the command tool again. With this change, as it likely more than 2 minutes ago or more than the maximum volume, that would mean I would need to reselect the entire building again as it also forces to forget the first corner I set

uecasm:
For that case, it would suffice to do the volume check only when you try to open the GUI, not when clicking the corners, similar to how it already does the anchor sanity check then.

And perhaps rather than forgetting the corners, it could just refuse to load the resource list and instead display a warning "scan box too large, are you sure you picked the right corners?" or something.

That sounds more like addressing the Real Problem™. 😁

PR #591, opened 2023-07-25
Add placeholder fill feature

someaddons:
- Adds a new feature to the scan tool, which automatically fills the top area of a schematic with placeholders to have the schematic integrated much better into the environment. E.g. it won't cut out huge rectangles when placed into a mountainside
- Allows players with a scan tool to pass through light placeholder blocks
- Made light placeholders semi transparent

Review please


uecasm:
Screenshot?

I'm not sure how useful a placeholder fill actually is.  Typically in level 1 you don't want to use placeholders, and in higher levels you need to carefully set specific blocks to air even when you want the majority to be placeholders.

Being able to pass through placeholders would be nice though.  Currently I use a two-step replace, where I swap the air to a unique block and then swap placeholders to air, then do the edits and then reverse this.  Occasionally this messes up other blocks (like converting grass/paths to dirt, or changing fence attachments) though.

src/main/java/com/ldtteam/structurize/blocks/schematic/BlockSubstitution.java
```diff
@@ -40,4 +42,15 @@ public VoxelShape getShape(final BlockState state, final BlockGetter worldIn, fi
     {
         return Shapes.box(.125D, .125D, .125D, .875D, .875D, .875D);
     }
+
+    @Override
+    public VoxelShape getCollisionShape(BlockState state, BlockGetter blockGetter, BlockPos blockPos, CollisionContext context) {
+        // Allow players to move through placeholders with a scantool in hand
+        if (context.isHoldingItem(ModItems.scanTool.get()))
```
uecasm: This only checks the item in the active hand; it might be nicer to check the offhand slot too (or even better, the whole hotbar).  Otherwise builders will keep getting stuck in the middle of placeholders when placing other blocks and have to keep swapping back.

someaddons: I kinda wanted to have it easily swappable, but offhand is a good idea, since it allows for both

someaddons:
image
image
Like for sure it is not perfect, thats also not the intention, but it'll make it easier to properly surrouned esp the air corners with placeholders to avoid buidings taking huge rectangle chunks out of the terrain

someaddons:
image

uecasm:
You might also want to write some usage docs for the changelog/wiki into the description.  I don't really see how the numbers in the first screenshot would produce the result in the second.

someaddons:
Usage doc is in tooltips on all the elements added to explain those

src/main/java/com/ldtteam/structurize/blocks/schematic/BlockSubstitution.java
```diff
@@ -40,4 +44,18 @@ public VoxelShape getShape(final BlockState state, final BlockGetter worldIn, fi
     {
         return Shapes.box(.125D, .125D, .125D, .875D, .875D, .875D);
     }
+
+    @Override
+    public VoxelShape getCollisionShape(BlockState state, BlockGetter blockGetter, BlockPos blockPos, CollisionContext context) {
+        // Allow players to move through placeholders with a scantool in hand
+        if (context instanceof EntityCollisionContext && ((EntityCollisionContext) context).getEntity() instanceof Player player)
```
Nightenom: Pattern instanceof

Nightenom (dismissed):

src/main/java/com/ldtteam/structurize/blueprints/v1/Blueprint.java
```diff
@@ -948,7 +953,15 @@ public boolean equals(final Object obj)
...
-        return Objects.equals(name, other.name) && palleteSize == other.palleteSize && getVolume() == other.getVolume() && Objects.equals(filePath, other.filePath);
+        return Objects.equals(name, other.name)
+                 && Objects.equals(fileName, other.fileName)
+                 && Objects.equals(filePath, other.filePath)
+                 && Objects.equals(packName, other.packName)
+                 && Objects.equals(rotation, other.rotation)
+                 && Objects.equals(mirror, other.mirror)
+                 && palleteSize == other.palleteSize
+                 && getVolume() == other.getVolume()
+                 && Objects.equals(renderSource, other.renderSource);
```
Nightenom: What is render source, this hash/equality controls render cache iirc, so it should ideally not create multiple buffers for same data

Edit: more like why is this method changed, what is the issue

someaddons: render source is the display position

Raycoms: Oh yeah, dont include the position then in here. We want to maximize caching.

Raycoms: I didnt add it intentionally

someaddons: There was a report on discord, about client lag caused by constant re-initialization of the renderer for a blueprint(saw it on a spark profiling they made). Couldnt reproduce it myself with just minecolonies, but since the hashcode was relatively weak it may have been possible that two different blueprints were fighting over the same renderer (was multiple ppl previewing things in an area, some existing building previews around aswell).

Since its kinda random, not easily reproducible I suspect the weak hashcode being the case, where I can see some rare cornercase of two blueprints producing the same hashcode, which is why I made it more robust/include more info

Raycoms: But equal hashcode will just display the same thing, even though its not the same thing. It shouldnt constantly switch, right?

Raycoms: (One of the players will see the wrong blueprint)

someaddons: they're calling update on each draw, after retriving the renderer via hashcode, so its possible that two handlers with different blueprints but matching hashcodes constantly re-inject themselves into the same renderer

src/main/java/com/ldtteam/structurize/client/BlueprintHandler.java
```diff
@@ -71,7 +71,6 @@ public void draw(final BlueprintPreviewData previewData, final BlockPos pos, fin
             rendererCache.put(blueprintHash, renderer);
         }
 
-        renderer.updateBlueprint(previewData);
```
Nightenom: Need git blame verification, will check tomorrow

src/main/java/com/ldtteam/structurize/blueprints/v1/Blueprint.java
uecasm: One of the laws of hashcodes is that any two things that are `equals` must have the same hash.  The hash is allowed to be weaker (though that needs justification -- "the field is mutable" is a valid justification though) but is not allowed to be stronger.

i.e. it's fine to have fields that don't participate in the equals, but then they must not participate in the hash.  This previously had that the wrong way around. (i.e. i.e. any field can be in both, only equals, or neither, but never only hash.)

It does sound like the render position should not be in the hash, though.  idk whether it should also not be in equals or not, I haven't looked at usage.

src/main/java/com/ldtteam/structurize/blueprints/v1/Blueprint.java
```diff
@@ -930,9 +930,13 @@ public int hashCode()
         final int prime = 31;
         int result = 1;
         result = prime * result + ((name == null) ? 0 : name.hashCode());
+        result = prime * result + ((fileName == null) ? 0 : fileName.hashCode());
```
Nightenom: Mergeable, if split away

src/main/java/com/ldtteam/structurize/client/BlueprintHandler.java
```diff
@@ -71,13 +74,25 @@ public void draw(final BlueprintPreviewData previewData, final BlockPos pos, fin
...
         Minecraft.getInstance().getProfiler().pop();
     }
 
+    /**
+     * Calcualtes a custom render hash for blueprints, including rotation and mirroring
+     * // TODO: No longer needed once rendering is able to rotate/mirror the displayed schematic
+     * @param blueprint
+     * @return
+     */
+    private static int getRenderHashFor(final Blueprint blueprint)
```
Nightenom: Split

src/main/java/com/ldtteam/structurize/client/BlueprintHandler.java
Nightenom: Split

Nightenom (changes requested):
Mergeable if hash things are split away, ideally fill issue/dm me with somewhat reproducible behaviour and I will try to fix it along with mirror/rotation pr next week

PR #602, opened 2023-08-07
Buildtool extension

someaddons:
- Adjust the buildtool slightly, adding optional unlock and confirmation callbacks

Review please


src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
```diff
@@ -18,31 +20,56 @@
...
 
 /**
  * BuildTool window.
  */
 public final class WindowExtendedBuildTool extends AbstractBlueprintManipulationWindow
 {
+    /**
+     * Default block requirement check
+     */
+    public static Predicate<Blueprint> BLOCK_BLUEPRINT_REQUIREMENT = blueprint ->
```
uecasm: Should also be `final`.

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
```diff
@@ -18,31 +20,56 @@
...
+    {
+        final BlockState anchor = blueprint.getBlockState(blueprint.getPrimaryBlockOffset());
+
+        if (anchor instanceof IRequirementsBlueprintAnchorBlock)
+        {
+            return (!((IRequirementsBlueprintAnchorBlock) anchor.getBlock()).areRequirementsMet(Minecraft.getInstance().level,
+              RenderingCache.getOrCreateBlueprintPreviewData("blueprint").getPos(),
```
uecasm, line 64: Pattern cast.

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
```diff
@@ -114,21 +151,59 @@ public enum ButtonType
...
+    public WindowExtendedBuildTool(
+      final BlockPos pos,
+      final int groundstyle,
+      @Nullable final BiConsumer<WindowExtendedBuildTool, Blueprint> selectionCallback,
+      @Nullable final Predicate<Blueprint> availableBlueprintPredicate)
     {
-        super(MOD_ID + BUILD_TOOL_RESOURCE_SUFFIX, pos, groundstyle,"blueprint");
+        super(MOD_ID + BUILD_TOOL_RESOURCE_SUFFIX, pos, groundstyle, "blueprint");
+        this.selectionCallback = selectionCallback;
+        this.availableBlueprintPredicate = availableBlueprintPredicate;
```
uecasm, line 185: I'm not sure if this is the case you're thinking of or a different one, but for Minecolonies I was thinking it would be useful just to extend the `IRequirementsBlueprintAnchorBlock` predicate rather than completely replace it, having a distinction between "this blueprint should be visible but not allow placement" and "this blueprint should show a warning but still allow placement" (and there being no problems, of course).  (Entirely hiding the blueprint is instead handled through `isInvisible`, which uses a separate predicate.)

In particular, for MineColonies, when you don't have the hut block it would still be nice to see the blueprint but not be able to place it; while not having the research but having the hut block would show a warning but still allow placement.

That would restore some pre-1.19 behaviour that's currently unavailable.  I'd imagine this would either mean a function that returns an enum, or multiple functions/predicates to check the different cases.

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
```diff
@@ -503,12 +585,14 @@ public void updateBlueprints(final List<Blueprint> inputBluePrints, final String
...
-               continue;
+                continue;
             }
 
             if (anchor.getBlock() instanceof ILeveledBlueprintAnchorBlock)
             {
-                final int level = ((ILeveledBlueprintAnchorBlock) anchor.getBlock()).getLevel(blueprint.getTileEntityData(RenderingCache.getOrCreateBlueprintPreviewData("blueprint").getPos(), blueprint.getPrimaryBlockOffset()));
+                final int level =
+                  ((ILeveledBlueprintAnchorBlock) anchor.getBlock()).getLevel(blueprint.getTileEntityData(RenderingCache.getOrCreateBlueprintPreviewData("blueprint").getPos(),
+                    blueprint.getPrimaryBlockOffset()));
```
uecasm, line 592: Pattern cast.

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
```diff
@@ -763,16 +849,19 @@ else if (buttonData.type == ButtonType.Blueprint)
...
             {
-                toolTip.addAll(((IRequirementsBlueprintAnchorBlock) anchor.getBlock()).getRequirements(Minecraft.getInstance().level, RenderingCache.getOrCreateBlueprintPreviewData("blueprint").getPos(), Minecraft.getInstance().player));
-                if (!((IRequirementsBlueprintAnchorBlock) anchor.getBlock()).areRequirementsMet(Minecraft.getInstance().level, RenderingCache.getOrCreateBlueprintPreviewData("blueprint").getPos(), Minecraft.getInstance().player))
+                if (anchor.getBlock() instanceof IRequirementsBlueprintAnchorBlock)
                 {
-                    PaneBuilders.tooltipBuilder().hoverPane(img).build().setText(toolTip);
-                    img.setImage(new ResourceLocation(MOD_ID, "textures/gui/buildtool/button_blueprint_disabled" + (hasAlts ? "_variant" : "") + ".png"), false);
-                    img.disable();
-                    return;
+                    toolTip.addAll(((IRequirementsBlueprintAnchorBlock) anchor.getBlock()).getRequirements(Minecraft.getInstance().level,
```
uecasm, line 853: Pattern cast.

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
Raycoms: old code, just auto formatter

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
```diff
@@ -763,16 +849,19 @@ else if (buttonData.type == ButtonType.Blueprint)
...
+                      RenderingCache.getOrCreateBlueprintPreviewData("blueprint").getPos(),
+                      Minecraft.getInstance().player));
                 }
+
+                PaneBuilders.tooltipBuilder().hoverPane(img).build().setText(toolTip);
```
Raycoms, line 858: Not sure what this change does

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
Raycoms: I dont think we should hide them.

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
someaddons: its old code, github just is bad at dealing with diffs if the parenthesis changes, new is just evaluating the unlock predicate instead of the previous check

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
someaddons: different usecase ye, I want to use it as schematic selector UI in scan features

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
Raycoms: But you removed an if and adjusted the tooltips

src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
uecasm: Was that intended for a different comment thread?

Raycoms (dismissed):

PR #605, opened 2023-08-17
Add setters to BlueprintPreviewData

Nightenom:
- Mirality wanted, probably needed for 1.19 but whatever 

Review please


uecasm (approved):
Looks good.  Though yes, 1.19 please too 😁 

Possibly of note is that it's not currently possible to set `packName` and `blueprintPath` except via deserialization, but my use case at least doesn't require those, since I'm using `setBlueprintFuture` directly.

Nightenom:
> Possibly of note is that it's not currently possible to set `packName` and `blueprintPath` except via deserialization, but my use case at least doesn't require those, since I'm using `setBlueprintFuture` directly.

These are apparently not meant to be public anyway, and probably even not meant to be fields too

PR #607, opened 2023-09-09
Add preview transparency

Nightenom:
- Images: [patreon discord](https://discord.com/channels/472875599422291968/472879361800929290/threads/1149988430915317770)
- Intentionally not with UI setting, this feature is alpha and breaks rendering sorting, but some ppl may use it
- I'm fine with UI settings (points at Ray to add it :smile:) as long as there is huge red text with warning. (btw it would be great if the current existing preview configs would be synced via forge client config)

Review please


uecasm:
Could we get it in a form where (without config file change) it's enabled by a property in `BlueprintPreviewData` (in particular so that it can be selectively enabled by things calling `StructureClientHandler`)?  Perhaps even a tuneable alpha value to allow variable transparency?

Bonus points if different `BlueprintPreviewData` can simultaneously request different alpha levels and it can honor that.

Nightenom:
Since it's hack that breaks previews transparency sorting even moreeeeeee I would say no, but can be added once renderer gets its rewrite

edit: willing to add selective disable at most now

Nightenom:
Finished, reviewable, add override if user already enabled transparency

UI warning + tooltip:
image

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -618,6 +626,45 @@ public BufferBuilder builder(RenderType renderType)
             }
             return buffer;
         }
+    }
 
+    public static class TransparencyHack
```
Raycoms, line 636: Please doc this here, case we someday might have to port this without you =D

src/main/java/com/ldtteam/structurize/client/gui/AbstractBlueprintManipulationWindow.java
```diff
@@ -412,22 +418,93 @@ public boolean isAllowedCharacter(final char c)
...
-                            Structurize.getConfig().set(typedSetting, newValue.intValue());
-                        }
-                        else if (oldValue instanceof Long)
+                        inputField.setTextColor(0xffe0e0e0); // vanilla defualt
+
+                        final DoubleValue rendererTransparency = Structurize.getConfig().getClient().rendererTransparency;
+                        if (setting == rendererTransparency && rendererTransparency.get() < 0)
                         {
-                            Structurize.getConfig().set(typedSetting, newValue.longValue());
+                            // TODO: move to standalone ui
```
Raycoms, line 426: do you want to address those todos before merging? =D

Nightenom: need blockui pr first :D plus its small window, can wait for porting, any other todos except this?

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
Nightenom: meh k :D

Nightenom: Done

PR #610, opened 2023-09-19
Allow blueprint handlers to become "mutually exclusive"

Thodor12:
- Blueprint handlers can be registered as mutually exclusive, this would mean only one of the handlers can be used simultaneously for placement.
- A use case for this is currently in Minecolonies, where we have 2 placement handlers for the plantation fields and the regular SurvivalHandler. The SurvivalHandler requires extra code to ensure there are no 2 "Assign to builder" buttons, this is because the SurvivalHandler can handle placement of anything, provided it is within colony borders.
Using this, that is not needed anymore, if the plantation field handler is defined first and the block is a plantation field, that one will be selected and the survival handler can then simply not be selected anymore.

Review please


Raycoms:
I dont like this. I think the canHandle methods should already handle this. I dont think one handler should be able to override all others.

PR #614, opened 2023-09-28
Back to the custom names

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/1156041864433107014)

- If the anchor block has a `name=Something` tag (directly on itself), this name will be used in the main blueprint list.
    - This is primarily intended to give names to decorations that would otherwise just appear as their (possibly numbered) filename, but it can also be used to override a building name if the designer feels that isn't appropriate for whatever reason.
    - If two blueprints that would otherwise be unrelated have the same name, then they will be grouped together as alternatives just as if they had the same building.
    - If two blueprints have the same building, but one has a tag and the other does not (or they have different tags) then they will be shown as separate top-level blueprints, not as alternates.
    - The displayed names for alternates are still just the blueprint filenames, as usual.
    - The tag only really matters for the level 1 blueprint, although it's recommended to use consistent tags for all levels.
    - There is currently no way to translate this name or to set a custom tooltip.
- When you click on a blueprint button that has alternatives and/or levels, previously the list of folders/blueprints at the bottom was completely cleared.  Now, a back button remains, allowing you to return to the list of blueprints that was visible previously, to allow quickly inspecting in and out of related buildings.

Review please (could port)


someaddons:
I dont think this makes sense as they are not translateable

uecasm:
> I dont think this makes sense as they are not translateable

Folder names and decorations aren't translatable already; this is not new.  I don't disagree that it would be nice to have a system to allow translation of all of those elements, but that would be a separate PR.

uecasm:
FWIW I have some code that will do some translation, which I can submit next, but I still do think it should be a separate PR.

uecasm:
Closing in favour of #643.

PR #638, opened 2023-12-20
Disables tagtool

someaddons:
-Disable tag tool useable in survival

Review please


Raycoms (dismissed):

src/main/java/com/ldtteam/structurize/network/messages/SetTagInTool.java
```diff
@@ -60,6 +61,12 @@ public void onExecute(final NetworkEvent.Context ctxIn, final boolean isLogicalS
             return;
         }
 
+        if (!ctxIn.getSender().isCreative())
+        {
+            ctxIn.getSender().displayClientMessage(Component.literal("Cannot use tag tool in survival"), false);
```
Nightenom: translation please

Nightenom (changes requested):

uecasm:
Some players rely on the tag tool being usable -- it can fix certain cases where things aren't tagged properly in the meantime before the schematic is actually fixed.

Whether it actually works or not depends on whether the tags are captured on block placement (will not work post-build) or are checked as needed (will work).

Also, the scan tool is deliberately usable in survival so that players can make decorations etc, and there's several reasons why they might want to tag their decorations.

Raycoms:
> Some players rely on the tag tool being usable -- it can fix certain cases where things aren't tagged properly in the meantime before the schematic is actually fixed.
> 
> Whether it actually works or not depends on whether the tags are captured on block placement (will not work post-build) or are checked as needed (will work).
> 
> Also, the scan tool is deliberately usable in survival so that players can make decorations etc, and there's several reasons why they might want to tag their decorations.

I thought about this yesterday too. For just "need fixing temporary" people can spawn it in creatively with the give command or creative menu. 

The scan tool is correctly usable in survivable and should stay. But being able to tag buildings while they are operating could allow people to potentially cheese things. E.g. infinitely extend the plantation field capacity.

Thodor12:
Well it doesn't work for plantation fields, I can guarantee you that much since field locations are only checked on placement time.
And even then there's a hard limit per field type, if there are additional ones those are ignored.

uecasm:
> For just "need fixing temporary" people can spawn it in creatively with the give command or creative menu.

Not if they're playing in survival on a server.  Or playing SP without cheats enabled.

> But being able to tag buildings while they are operating could allow people to potentially cheese things.

Which things?  As Thodor said the plantation has built-in limits even if you add tags (and it captures tag on block placement so adding them later does nothing anyway) -- that was intended so that style authors can't break game balance either.  `invisible` and `leisure` tags will work after placement, but aren't cheesable (and are good examples of ones that players without creative might legitimately want to use in decorations).  I don't think any of the other tags are cheesable either.

someaddons:
Players simply should not be able to mess with internal data where they can easily break stuff

uecasm:
> Players simply should not be able to mess with internal data where they can easily break stuff

Despite all examples being beneficial and no examples breaking anything?  I don't agree.  Besides, if players do break things, they get to keep the pieces.  (i.e. given that it is more helpful than harmful, it is actively harmful to disable the feature, with no upside.)

Raycoms:
> > Players simply should not be able to mess with internal data where they can easily break stuff
> 
> Despite all examples being beneficial and no examples breaking anything? I don't agree. Besides, if players do break things, they get to keep the pieces. (i.e. given that it is more helpful than harmful, it is actively harmful to disable the feature, with no upside.)

I dont think a debug tool being available for crafting makes a lot of sense. People look up obtainable stuff in survival in JEI for mods and play around with it.  Its confusing for players at the least.

And, instead of telling people to fix the tag, we should fix the tag and update the mod =D

PR #640, opened 2023-12-27
Fix fog setup inside level render events

Nightenom:
- fix changes in fog setup when inside non-AFTER_LEVEL renderWorldEvents

Review please

### DIFF
--- a/src/api/java/com/ldtteam/structurize/api/util/ItemStackUtils.java
+++ b/src/api/java/com/ldtteam/structurize/api/util/ItemStackUtils.java
@@ -192,7 +192,6 @@ public final class ItemStackUtils
      * @param itemStack2 The right stack to compare.
      * @return True when they are equal except the stacksize, false when not.
      */
-    @NotNull
     public static boolean compareItemStacksIgnoreStackSize(final ItemStack itemStack1, final ItemStack itemStack2)
     {
         return compareItemStacksIgnoreStackSize(itemStack1, itemStack2, true, true);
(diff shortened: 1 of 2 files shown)


Raycoms (approved):

marchermans (approved):

uecasm (approved):
This does seem to fix it.  Although that begs the question of whether it's intended that it redraws everything at all the render stages.  Isn't it more typical to only draw at one stage?  Or to draw specific things at specific stages?

uecasm:
Hmm.  Actually, it does only get called at one stage, so this code change has the effect of never calling that fog code at all, as far as I can tell.

In Structurlze, `ClientEventSubscriber` calls `draw` only at one of `Stage.AFTER_CUTOUT_MIPPED_BLOCKS_BLOCKS` or `Stage.AFTER_TRANSLUCENT_BLOCKS` depending on the translucency setting.

In MineColonies, `WorldEventContext` calls `draw` only at `Stage.AFTER_CUTOUT_MIPPED_BLOCKS_BLOCKS`.  (Possibly it needs to similarly check the setting?)

So there is no case when the `if` blocks would be entered. which suggests either that the code could just be deleted entirely, or this isn't the intended fix.

uecasm (changes requested):
Per above, this doesn't seem quite right after all.

Nightenom:
Yeah, I know code I made :D end-users may currently render blueprints possibly at any stage (this renderer replicates entire LevelRenderer#renderLevel(...) method up to the most reproducible point, so it doesnt really care from which context it has been called)

PR #643, opened 2024-01-12
Back to the custom names 1.20

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/1156041864433107014)

- If the anchor block has a `name=Something` tag (directly on itself), this name will be used in place of the blueprint's filename.
    - This is primarily intended to give nicer names to decorations that would otherwise just appear as their (possibly numbered) filename, but it can also be used to override the "alternates" name shown for MineColonies buildings.
    - The tag must be used consistently at all levels of the blueprint, otherwise they will "split".
    - There is currently no way to translate this name, but that's not different from existing behaviour.  A translation PR will follow.
- When you click on a blueprint button that has alternatives and/or levels, previously the list of folders/blueprints at the bottom was completely cleared.  Now, a back button remains, allowing you to return to the list of blueprints that was visible previously, to allow quickly inspecting in and out of related buildings.
- The `isInvisible` check has been extracted to a utility method since MineColonies wants to be able to call it too.

Review please (could port)

This is a port of #614 but it has a different implementation that doesn't override the building name.

### DIFF
--- a/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintTagUtils.java
+++ b/src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintTagUtils.java
@@ -2,15 +2,19 @@ package com.ldtteam.structurize.blueprints.v1;
 
 import com.ldtteam.structurize.blockentities.interfaces.IBlueprintDataProviderBE;
 
+import com.ldtteam.structurize.blocks.interfaces.IInvisibleBlueprintAnchorBlock;
+import com.ldtteam.structurize.util.BlockInfo;
 import net.minecraft.core.BlockPos;
 import net.minecraft.nbt.CompoundTag;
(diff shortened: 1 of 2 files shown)


Raycoms (approved):

uecasm:
Not really sure what the difference is between `1.20.1` and `main` or why the base branch was changed.  But what else is needed to merge this?

Thodor12:
`version/main` is now where the porting to 1.20.4 is happening

uecasm:
> `version/main` is now where the porting to 1.20.4 is happening

Not according to the commit history, it isn't.

Nightenom:
It stems from miscommunication between Ray and me, when I thought we agreed on making main stale during porting to mitigate conflicts in porting branch (and also mark where did the porting split happen to later merge all unported commits) - this was applied successfully on blockui, not so successfully here..

PR #648, opened 2024-01-20
Handle dripstone

Raycoms:
Closes #
Closes #
Closes #

- Proper handling of dripstone
-
-

Review please


Nightenom (approved):
Ok, so I guess handling it via chain placing should work, I'm not sure if the placeDripstoneInDir shouldn't actually be while cycle instead

PR #649, opened 2024-01-24
Qol

Raycoms:
Closes #
Closes #
Closes #

- Add ding sound to scan tool clicks
- Allow remove filtered in scan tool (also move some things around to  make space)
- Add to undo/redo window more info
- Increase default undo/redo cache to 50
- Add entity removal step to structure placement handling (maintain backward compat)

image

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/util/ChangeStorage.java
+++ b/src/main/java/com/ldtteam/structurize/util/ChangeStorage.java
@@ -2,6 +2,7 @@ package com.ldtteam.structurize.util;
 
 import com.ldtteam.structurize.Structurize;
 import com.ldtteam.structurize.management.Manager;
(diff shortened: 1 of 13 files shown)


src/main/java/com/ldtteam/structurize/util/ChangeStorage.java
```diff
@@ -46,7 +47,7 @@ public class ChangeStorage
     /**
      * The operation which was done
      */
-    private final String operation;
```
Nightenom, line 49: Uff, do we really want logic based on components? or is this purely UI (could be better variable name then)?

Raycoms: Purely used for UI at this point.

someaddons (approved):

PR #652, opened 2024-02-06
Change display entity to only paste in schematic mode

someaddons:
Closes #
Closes #
Closes #

- Fix entity detection in scan tool
- Change display entity to only paste in schematic mode

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
+++ b/src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
@@ -12,13 +12,17 @@ import com.ldtteam.structurize.util.BlockUtils;
 import com.ldtteam.structurize.util.ChangeStorage;
 import net.minecraft.core.BlockPos;
 import net.minecraft.nbt.CompoundTag;
+import net.minecraft.world.entity.Display;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.entity.Mob;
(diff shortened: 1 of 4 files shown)


Raycoms (dismissed):

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -439,6 +443,11 @@ public BlockPlacementResult handleEntitySpawn(
                                 continue;
                             }
 
+                            if (entity instanceof Display.TextDisplay && (!handler.isCreative() || handler.fancyPlacement()))
```
Nightenom, line 446: agree, but why only text display?

Raycoms: Just thinking about it. Don't we have a similar check somewhere about LivingEntities already? Could add TextDisplay to the list there

someaddons: since possibly you could do visual stuff with block/item display didnt want to block those inherently

someaddons: > Just thinking about it. Don't we have a similar check somewhere about LivingEntities already? Could add TextDisplay to the list there
havent seen anything like that :D

Nightenom (approved):

uecasm:
Any plans to backport this to 1.19?  I currently have one set of blueprints for both 1.19 and 1.20.

PR #662, opened 2024-04-10
easier scan name, allow double for version (handle double internally)

Raycoms:
Closes #
Closes #
Closes #

- easier scan name
-  allow double for version (handle double internally)
- show full name in tooltip
image

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
+++ b/src/main/java/com/ldtteam/structurize/client/gui/WindowExtendedBuildTool.java
@@ -844,6 +844,7 @@ public final class WindowExtendedBuildTool extends AbstractBlueprintManipulation
             {
                 final String name = getCustomName(firstBlueprint, firstBlueprint.getFileName());
                 img.setText(Component.literal(name));
+                toolTip.add(Component.literal(name));
             }
             img.setVisible(true);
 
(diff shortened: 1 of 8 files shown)


Nightenom (commented):
Why double? doesn't seem the best for versioning - or is it only for visual things?

Raycoms:
> Why double? doesn't seem the best for versioning - or is it only for visual things?

1.0 1.1 1.2 1.3 instead of only 1 2 3 4

Nightenom:
ooof, imho use sth better than double, anything like record, artifact version, or idk what's available
double is based on locale etc.

uecasm:
`ArtifactVersion` sounds like a probably good choice.  The parser could even be smart enough to support both plain int and stringified artifact version, for backwards compatibility.

Raycoms:
> `ArtifactVersion` sounds like a probably good choice. The parser could even be smart enough to support both plain int and stringified artifact version, for backwards compatibility.

The question is pretty much. Do we need it? Double supports both an int and a double (e.g. 1 and 1.0 and 1.2 and 2,3,4).

Artifact version supports that too, but in this context is an overkill compared to what we need. Like, int versions are technically enough, but are a bit boring if you want to show the progress properly.

uecasm:
If you had to write your own string parser, I'd say no.  Since there's one you can use for free, I say why not? 😀 

With a full artifact version it might even make sense to display it (in tiny font) on the style pack selector, so people can see if they have a BETA pack.  Or you could get a more descriptive warning message if there's a mismatch between client/server.

> double is based on locale etc.

Though fwiw that's not really a problem here.  The JSON parse is locale-independent and currently it's not displayed anywhere, so there's no other locale-related problems.  There might be precision-related problems, but that's unlikely to be an issue in practice.

someaddons (approved):

PR #665, opened 2024-04-19
Add blacklist

Raycoms:
Closes #
Closes #
Closes #

-
- Add block blacklist, to allow serverowners and packmakers to exclude blocks that allow duping
-

Review please


someaddons (approved):

src/main/java/com/ldtteam/structurize/blueprints/v1/BlueprintUtil.java
```diff
@@ -74,6 +75,10 @@ public static Blueprint createBlueprint(
         for (final BlockPos mutablePos : BlockPos.betweenClosed(pos, pos.offset(sizeX - 1, sizeY - 1, sizeZ - 1)))
         {
             BlockState state = world.getBlockState(mutablePos);
+            if (state.is(ModTags.BLUEPRINT_BLACKLIST))
```
Nightenom, line 78: uhh what? what's the code path of this method? we definitely want to save what we scan fully, such blacklisting should take place during structure placing

Raycoms: I'm doing both. Preventing people to scan blacklisted blocks (also to make it harder to try to bypass these things in past versions. Gotta scan in past and place in past!)

Nightenom: This is hella flawed imho, especially you dont care about what you scan/place in creative, also you usually don't care if you play singleplayer (regardless of pack maker blacklisting or not), makes really no sense for me to do this during scanning

Raycoms: This is not only for exploity stuff, but also for stuff that crashes, etc (doesn't work well at all when scanned).

PR #671, opened 2024-05-29
no z fighting

someaddons:
Closes #
Closes #
Closes #

-
-
-

Review please


Nightenom (commented):
I would make it at least config, also this might not work in higher numbers but w/e

someaddons:
seems to work fine at 500k 500k out still, though interestingly I found a different issue, being so far out causes the preview to wobble with player rotation(even without this change)

Raycoms:
On newest struct?

someaddons:
ye

Nightenom:
> Member Author

Nightenom:
> seems to work fine at 500k 500k out still, though interestingly I found a different issue, being so far out causes the preview to wobble with player rotation(even without this change)

whoopsie close.. yeah you've reported this issue earlier this year, havent got to it yet

someaddons:
ooh you're right, was when I had a coordinate issue with tavern sitting entities aswell :D

PR #718, opened 2025-02-04
Try to reduce serverlag a bit

Nightenom:
- apparently `isOldChunkAround` has some chunk worker IO thingy inside, we know where we are here though, especially we expect that a chunk, in which the structure is being built, is already fully loaded including other chunks in worldgen range

## Testing
- [ ] Yes I tested this before submitting it. Nope
- [ ] I also did a multiplayer test. Nope

Review please


someaddons (approved):

Nightenom:
Blind merging, change should be relatively stable

PR #726, opened 2025-03-18
Placeholder preview setting

someaddons:
- Allow Blueprint preview data to select if placeholders are displayed

## Testing

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
+++ b/src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
@@ -1,5 +1,6 @@
 package com.ldtteam.structurize.client;
 
+import com.ldtteam.structurize.Structurize;
 import com.ldtteam.structurize.blockentities.BlockEntityTagSubstitution;
 import com.ldtteam.structurize.blocks.ModBlocks;
 import com.ldtteam.structurize.blueprints.v1.Blueprint;
@@ -689,11 +690,17 @@ public class BlueprintRenderer implements AutoCloseable
                 return;
             }
 
-            if (overrideValue < 0 || overrideValue > THRESHOLD)
+            float alpha = Structurize.getConfig().getClient().rendererTransparency.get().floatValue();
+            if (overrideValue != -1)
+            {
+                alpha = overrideValue;
+            }
+
+            if (alpha < 0 || alpha > THRESHOLD)
             {
                 return;
             }
-            final float alpha = Mth.clamp(overrideValue, 0, 1);
+
             applied = true;
 
             RenderSystem.enableBlend();
--- a/src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
+++ b/src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
@@ -1,6 +1,5 @@
 package com.ldtteam.structurize.client;
 
-import com.ldtteam.structurize.Structurize;
 import com.ldtteam.structurize.blockentities.BlockEntityTagSubstitution;
 import com.ldtteam.structurize.blocks.ModBlocks;
 import com.ldtteam.structurize.blueprints.v1.Blueprint;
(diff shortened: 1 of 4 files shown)


Raycoms (dismissed):

src/main/java/com/ldtteam/structurize/client/fakelevel/BlueprintBlockAccess.java
```diff
@@ -40,7 +41,12 @@ public BlockState getBlockState(final BlockPos pos)
 
     public BlockState prepareBlockStateForRendering(final BlockState state, final BlockPos pos)
     {
-        if (Structurize.getConfig().getClient().renderPlaceholdersNice.get())
+        return prepareBlockStateForRendering(state, pos, null);
+    }
+
+    public BlockState prepareBlockStateForRendering(final BlockState state, final BlockPos pos, final BlueprintPreviewData previewData)
+    {
+        if (previewData == null && Structurize.getConfig().getClient().renderPlaceholdersNice.get() || (previewData != null && previewData.getRenderBlocksNice()))
```
Nightenom: imho could be simplified with ternary

Nightenom (commented):
approveable - just one question: how is the setter handled in terms of reseting the cached renderer? either please provide javadoc with what to call along or handle the logic code-wise

someaddons:
The cache cleanup is on the caller similar to the other places using it, just like the other changes as rotation, server sync etc are per preview instance. From the minecolonies side we mostly use our own cache there and call the renderer directly, from the looks it seems the structurize cache is soley used for the buildtool rendering

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -690,13 +689,11 @@ public static void apply(final float overrideValue)
                 return;
             }
 
-            float alpha = Structurize.getConfig().getClient().rendererTransparency.get().floatValue();
```
Nightenom, line 693: This won't work, this is dynamic value and needs to be reliazed in this class. Please revert (it might be possible to move this out, but it would be a rather large refactor to keep the current logic)
Also for this reason there is the `getOverridePreviewTransparency` so if you need some more logic please enhance that
And if you plan to use this in some real scenarios, then it was not built for that, this transparency hack still might result in potential x-rays/artefacts etc. That's why this was done as hard disable if disabled in config. If we want real transparency for previews then we need much more work

Nightenom (changes requested):
Afaik BlueprintRenderer is not directly referenced in mcol, and definitely should not be. Fine with the explanation though.
Nevertheless the recent changes are a no from my POV

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
someaddons: I can add the config value back again, but the override will need to take precedence, else the override does not work unless the user manually enabled transparency in the config before. I dont care about artifacts/sometimes broken rendering

Nightenom: Yeah, but I care, it's called hack for a reason, and if you want to use this as default for your feature then consider using proper transparency specifically for that
If you really want to use this for w/e, then displaying the same huge warning message (as we do in build tool) is fine with me
_Else you can straight flag this mod with "hack to see through all transparency blocks - use it via preview"_

Plus you left out the clamp call, which is a must unless you want disco

Raycoms: tbh, it has been working fairly well so far, so I wouldn't be worrying too much about it. Not like it's a production competitive IEEE title. I don't even think proper transparency works in minecraft =D Not aware of anybody that does transparency things that achieved it =D

Raycoms: IEEE, lol, I mean AAA =D either way clamp is missing then we can do shipit.

PR #734, opened 2025-04-19
Extensible handlers for entities

uecasm:
Closes [discord request](https://discord.com/channels/139070364159311872/724810039281451028/1362602381707186297)
Closes [discord request](https://discord.com/channels/139070364159311872/724810019928801290/1257558255379415040)

- Moves the "Remove Filtered" button below the resource list to hopefully reduce misclicks.
- Fixes some off-by-one errors when converting bounding boxes to `AABB`s (particularly for entity searches).
- Adds "entity handlers", which similar to placement handlers are extensible by addon mods to hook into UI and placement behaviour, instead of relying on hard-coded checks.
    - As before, only certain entities will place in survival mode (assuming Minecolonies or another survival handler is installed).
    - Entities without any handler are not placeable in survival (but can still be placed in creative).  When a handler is defined then the entity becomes placeable by default, unless the handler says otherwise.
    - Glow item frames now require themselves as resources, instead of being a free upgrade from normal item frames.
- The scan tool entity list now shows some extra icons:
    - Up to four item icons showing the resource types that will be required in survival mode.
    - A red/green indicator showing whether this entity can be placed in survival mode.  (This is somewhat irrelevant to Structurize alone but useful if there's any survival mod on top.)

image

## Testing

Review please (should port)


src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Gets the adjusted position in the world for the given entity.
+     *
+     * @param entity the entity.
+     * @param pos    the anchor position in world.
+     * @return       the adjusted entity position.
+     */
+    default Vec3 adjustPosition(final Entity entity, final BlockPos pos)
```
Nightenom: I would suggest using longer name than just pos, to emphasize that this is anchor or w/e it actually is
```suggestion
    default Vec3 adjustPosition(final Entity entity, final BlockPos worldRootPos)

```

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!canHandle(entity)) return ActionProcessingResult.PASS;
+        if (!canPlace(entity, handler.isCreative(), handler.fancyPlacement())) return ActionProcessingResult.DENY;
+
+        Vec3 posInWorld = adjustPosition(entity, pos);
+        entity.moveTo(posInWorld.x, posInWorld.y, posInWorld.z);
+
+        final List<? extends Entity> list = entity.level().getEntitiesOfClass(entity.getClass(), AABB.unitCubeFromLowerCorner(posInWorld));
+        for (Entity worldEntity : list)
+        {
+            if (worldEntity.position().equals(posInWorld))
```
Nightenom, line 75: This seems sus, are we considering same class same pos as same entities? what about their content/wouldn't it be better to use uuid to prevent dupes? (along with consideration of vanilla in world uuid deduplication)

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Method used to get the required items to place an entity.
+     *
+     * @param entity  the entity being placed.
+     * @return        the list of items.
+     */
+    default List<ItemStack> getRequiredItems(final Entity entity)
+    {
+        return List.of(entity.getPickedResult(new HitResult(entity.position()) {
```
Nightenom: Could we use normal entity hit result with all the data?

Nightenom (changes requested):
will do full review later, just some quick things to consider

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
uecasm: This is just exactly what it did before.  In some cases (e.g. client side) it's called with a fresh entity rather than the real entity, and I haven't seen any case where it actually cared about the hit result anyway.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
uecasm: This is essentially what it did before, except it used an even larger range of blocks to search within (possibly to allow for mob wandering slightly?).

It can't use UUID because it's guaranteed that the entity in-world will have a different UUID from the entity in-blueprint.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
uecasm: Fixed

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
uecasm: Can't see any particular reason not to use `EntityHitResult` though.

src/main/java/com/ldtteam/structurize/client/gui/WindowScan.java
```diff
@@ -559,23 +567,38 @@ public int getElementCount()
...
+                                .values().stream().sorted(Comparator.comparingInt(ItemStorage::getAmount).reversed())
+                                .toList();
+                        break;
+                    }
                 }
-                else if (entity == EntityType.MINECART)
+                rowPane.findPaneOfTypeByID(RESOURCE_QUANTITY_MISSING, Text.class).setText(Component.literal(Integer.toString(entities.size())));
+                rowPane.findPaneOfTypeByID(RESOURCE_ICON, ItemIcon.class).setItem(entityIcon);
+                rowPane.findPaneOfTypeByID(RESOURCE_NAME, Text.class).setText(entityType.getDescription());
+                for (int i = 1; i <= 4; ++i)
```
Nightenom: nitpick - me basically don't care, it's just more common to use this when accessing arrays/lists
```suggestion
                for (int i = 0; i < 4; ++i)
```

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Handler for all kinds of mob
+     */
+    public static class MobEntityHandler implements IEntityHandler
+    {
+        @Override
+        public boolean canHandle(final Entity entity)
+        {
+            return entity instanceof Mob;
```
Nightenom, line 86: don't have ide open, what is lower livingEntity or mob in entity class hierarchy? iirc mob implements livingentity, if so why not instanceof livingEntity?
(i see it was in old code, just asking)

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
```diff
@@ -0,0 +1,226 @@
...
+        @Override
+        public boolean canHandle(final Entity entity)
+        {
+            return entity instanceof Display.TextDisplay;
+        }
+
+        @Override
+        public boolean canPlace(final Entity entity, final boolean isCreative, final boolean isFancy)
+        {
+            return isCreative && !isFancy;
```
Nightenom, line 117: wonder that all Display types should be placed if `isCreative`, if mcol wants `!isFancy` then it should be there

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
```diff
@@ -0,0 +1,226 @@
...
+        {
+            return entity instanceof ItemFrame;
+        }
+
+        @Override
+        public List<ItemStack> getRequiredItems(final Entity entity)
+        {
+            final List<ItemStack> request = new ArrayList<>();
+            request.add(((ItemFrame) entity).getItem().copyWithCount(1));
+            request.add(new ItemStack(Items.ITEM_FRAME));
```
Nightenom: you are missing Items.GLOW_ITEM_FRAME

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
Nightenom: note for myself to revisit this in testing

Nightenom (changes requested):
I hope you tested thoroughly, I personally use following (4 years old :D) entity blueprint along with one for paintings, both in also mirror settings. I want to do in-game testing myself, but not sure when will get to it

it's txt because github uploading.. before using rename to .blueprint
[debug_entities.txt](https://github.com/user-attachments/files/19964841/debug_entities.txt)

uecasm:
> testing

image

You can see most of the test blueprints I was using in the description screenshot.  Mostly paintings and item frames, but I tried several other things too.  Pictured here is your blueprint rotated twice, placed, and scan-boxed again, it looks ok AFAICT.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: Mob is derived from LivingEntity.  There's probably no particular reason this couldn't be LivingEntity, but I was trying to keep previous behaviour, and that's what it did before.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: This also matches previous behaviour.  But even for pure structurize it makes sense to me to keep them around only for "schematic paste" and not "constructed", because the purpose of these was to show notes in the preview that don't survive when built.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: This also matches prior behaviour.  Both ITEM_FRAME and GLOW_ITEM_FRAME are matched by this handler but it only demands ITEM_FRAME as a resource.  The glowing is free.  This matches how signs work (since they don't have a glowing item version).

image

src/main/java/com/ldtteam/structurize/client/gui/WindowScan.java
uecasm: That's fair.  It seemed simpler to do it this way to keep the ids 1-based, but I don't mind changing this.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
Raycoms: We have been including glow itemframes as costs in Schematics, we probably should match that indep

uecasm: I assume you meant "haven't".  Because we haven't.

https://github.com/ldtteam/Structurize/blob/a6e5b5e1147c1eb08eedc4316d8441e287cc6cb2/src/api/java/com/ldtteam/structurize/api/util/ItemStackUtils.java#L162-L171

This is the code being replaced by this handler.  As you can see, it matches both kinds of item frame but reports only the basic frame as required resources.  This was deliberate because it's how signs also work.  (Glowing signs are purely a blockstate, not a persistent item.)

Raycoms: Yes, that stems from pre glowing one still. For signs I think there is no glowing sign item right? So there it's fine, but for itemframes its craftable, no?

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: `SpawnEggItem(EntityType<? extends Mob> p_43207_, ...`

Actually this is probably why.  It's technically only legal to create spawn eggs for Mobs.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: Yeah.  I mean we could add a cost for glowing signs as well if we wanted to, with a placement handler.  I assumed it was a deliberate choice not to do so for both (at least I'm pretty sure Moobien and Train said they were happy it was "free").

MotionlessTrain: I can’t recall I said that. But then again, I can’t recall the recipe for glowing item frames
If it is just a glowstone and an item frame, personally I think it’s fine

uecasm: It requires a glowsquid ink sac (which are fairly rare).  This is also the requirement to make a sign have glowing text.  Which means glowing stuff would be a L3+ item only, most likely -- and people wanted to use them at L1.

uecasm: Technically deep ocean stuff is L4+.  Though they can also be found in cave pools.  Pretty rare there too.

Raycoms: I find them in a lot of large caves, when there is a little waterfall they spawn and die on their own

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
Nightenom: Makes sense, wondering if we want to resolve the rest of the set - ie. complement to Mob in LivingEntity set

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
Nightenom: I see that, for now fine, but later for in-blueprint notes we should rather create our extension of Diplay.Text that is backed using tags.
Remaining question: the other Display types? (btw I get you tried to keep up with old code, but since we're unifying logic here we could make it more proper right away)

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
Nightenom: https://github.com/ldtteam/Structurize/blob/a6e5b5e1147c1eb08eedc4316d8441e287cc6cb2/src/main/java/com/ldtteam/structurize/client/gui/WindowScan.java#L564-L567

Just saying this was there in code you've removed :D

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: Any entity not handled by any entity handler will still get default behaviour: placeable in creative mode only, no resources listed, best guess at icon.

The only thing the Mob handler really does differently from default is to give it a resource list, which is kinda redundant since it can't be placed in survival anyway.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: Yeah, we could extend support for those too.

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: > Just saying this was there in code you've removed :D

Yes, but that was the icon, not the resources.  If you look at the screenshot above you can see it still keeps the glow item frame icon, because that's already the default pickblock behaviour.

uecasm: Glow item frames are now required resources, instead of a free upgrade.

Raycoms (approved):
This looks good. I think as a next step we should maybe look at better entity matching.
(I.e. not only same pos and type, but also same inventory) and then do a "replace" (break and pick up).

Nightenom:
I've DMed you @uecasm, I'm generally fine with changes but I would like to test it myself on 1.21, so either you or I need to port this branch first, tell me whatever works best for you

PR #736, opened 2025-04-27
Placeholders #2

someaddons:
Closes #
Closes #

- Now properly: Allows previewdata to determine solid placeholder replacement :)
-

## Testing

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
+++ b/src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
@@ -117,6 +117,11 @@ public class BlueprintRenderer implements AutoCloseable
         blockAccess.setBlockEntities(tileEntitiesMap);
         blockAccess.setEntities(entities);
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -143,6 +117,11 @@ private void init(final Blueprint blueprint, final Map<Object, Exception> suppre
         blockAccess.setBlockEntities(tileEntitiesMap);
         blockAccess.setEntities(entities);
 
+        if (previewData.getSolidSubstitutionOverride() != null)
+        {
+            blockAccess.setSolidSubstitutionOverride(previewData.getSolidSubstitutionOverride());
```
Nightenom: not bad, could we unset this after in the end of the init method?

someaddons: oh right for the other previewdata's which may have not set it

Nightenom (approved):

PR #737, opened 2025-04-29
Port726

someaddons:
Closes #
Closes #

- port of #726
-

## Testing

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/client/fakelevel/BlueprintBlockAccess.java
+++ b/src/main/java/com/ldtteam/structurize/client/fakelevel/BlueprintBlockAccess.java
@@ -7,6 +7,7 @@ import com.ldtteam.structurize.Structurize;
 import com.ldtteam.structurize.blockentities.BlockEntityTagSubstitution;
 import com.ldtteam.structurize.blocks.ModBlocks;
 import com.ldtteam.structurize.blueprints.v1.Blueprint;
+import com.ldtteam.structurize.storage.rendering.types.BlueprintPreviewData;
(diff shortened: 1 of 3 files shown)


src/main/java/com/ldtteam/structurize/client/fakelevel/BlueprintBlockAccess.java
```diff
@@ -47,7 +48,12 @@ public BlockState getBlockState(final BlockPos pos)
 
     public BlockState prepareBlockStateForRendering(final BlockState state, final BlockPos pos)
     {
-        if (Structurize.getConfig().getClient().renderPlaceholdersNice.get())
+        return prepareBlockStateForRendering(state, pos, null);
+    }
+
+    public BlockState prepareBlockStateForRendering(final BlockState state, final BlockPos pos, final BlueprintPreviewData previewData)
```
Nightenom, line 54: Huh? Did i actually approved that in prev PR? the getBlockState(pos) must return correct blockState in as most as possible contextes - this should be done in similar way as `solidSubstitutionOverride`

Nightenom (changes requested):
Sorry I missed that earlier, could you please fix following also the original PR?

src/main/java/com/ldtteam/structurize/client/fakelevel/BlueprintBlockAccess.java
someaddons: ye saw that too, I'll make it in an extra PR so I can merge that one to both versions

Raycoms (approved):

Nightenom (approved):

PR #746, opened 2025-05-21
Adjust render nice check

someaddons:
Closes #
Closes #

- Change rendering nice check to be context insensitive similar to the placeholder override

## Testing
- [ x] Yes I tested this before submitting it.

Review please


src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -194,6 +195,7 @@ private void init(final BlueprintPreviewData previewData, final Map<Object, Exce
         }
 
         blockAccess.setSolidSubstitutionOverride(null);
+        blockAccess.setRenderBlocksNiceOverride(Structurize.getConfig().getClient() != null && Structurize.getConfig().getClient().renderPlaceholdersNice.get());
```
Nightenom: This is client only code, same below

```suggestion
        blockAccess.setRenderBlocksNiceOverride(Structurize.getConfig().getClient().renderPlaceholdersNice.get());
```

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
```diff
@@ -194,6 +195,7 @@ private void init(final BlueprintPreviewData previewData, final Map<Object, Exce
         }
 
         blockAccess.setSolidSubstitutionOverride(null);
+        blockAccess.setRenderBlocksNiceOverride(Structurize.getConfig().getClient().renderPlaceholdersNice.get());
```
Raycoms, line 199: I think here too

someaddons: think you looked at the outdated one? I merged the removal of the null check earlier :D

MotionlessTrain: That is actually the same line Nightenom commented about earlier :D

Raycoms: So we have a client config not null check here before this?

someaddons: yes because this is in the renderer, that was what nightenom was saying its not needed to check on the client side

Raycoms: ahhh

Raycoms (approved):

src/main/java/com/ldtteam/structurize/client/BlueprintRenderer.java
Nightenom: It's still in the other file, mergeable

PR #753, opened 2025-06-04
Entity handlers for 1.21

uecasm:
- Port #734
    - Omits the off-by-one AABB fixes, because 1.21 already fixed this differently.

## Testing

Review please (do not port)


src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -367,51 +369,28 @@ public BlockPlacementResult handleEntitySpawn(
...
-                                    continue;
-                                }
-
-                                if (simulate || !this.handler.hasRequiredItems(requiredItems))
-                                {
-                                    return new BlockPlacementResult(worldPos, BlockPlacementResult.Result.MISSING_ITEMS, requiredItems);
-                                }
-                            }
-                            else if (requiredItems == null)
+                            if (!handler.isCreative() && !handler.hasRequiredItems(requiredItems))
```
Raycoms, line 391: No simulate flag check here?

src/main/java/com/ldtteam/structurize/placement/handlers/entity/IEntityHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        if (!canHandle(entity)) return ActionProcessingResult.PASS;
+        if (!canPlace(entity, handler.isCreative(), handler.fancyPlacement())) return ActionProcessingResult.DENY;
+
+        Vec3 posInWorld = adjustPosition(entity, zeroPos);
+        entity.moveTo(posInWorld.x, posInWorld.y, posInWorld.z);
+
+        final List<? extends Entity> list = entity.level().getEntitiesOfClass(entity.getClass(), AABB.unitCubeFromLowerCorner(posInWorld));
+        for (Entity worldEntity : list)
+        {
+            if (worldEntity.position().equals(posInWorld))
```
Raycoms, line 76: I'd like to add a check here too where we make sure the content is the same too. We should test with itemframes with upgrading contents then.

PR #756, opened 2025-06-08
Random fixes for Mirality to try

Nightenom:
- gradle wrapper will disappear if you update from main
- fixes PAINting placement once and forever (?)
- correctly process entity content in neo caps
- reformat handleEntitySpawn to not have 16565179 tabs


src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
```diff
@@ -195,14 +179,6 @@ public boolean canHandle(final Entity entity)
         {
             return entity instanceof ContainerEntity;
         }
-
-        @Override
-        public List<ItemStack> getRequiredItems(final Entity entity)
-        {
-            final List<ItemStack> request = new ArrayList<>(IEntityHandler.super.getRequiredItems(entity));
-            request.addAll(((ContainerEntity) entity).getItemStacks());
```
Raycoms, line 203: Why rm these here?

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
```diff
@@ -350,67 +344,67 @@ public BlockPlacementResult handleEntitySpawn(
...
+
+                List<ItemStack> requiredItems = List.of();
+                nextHandler:
+                for (final IEntityHandler entityHandler : EntityHandlers.handlers)
+                {
+                    switch (entityHandler.checkPlacement(handler, entity, zeroPos))
+                    {
+                        case DENY:
+                            continue nextEntity;
+                        case PASS:
```
Raycoms, line 380: Are you sure thr logic here is correct ?

src/main/java/com/ldtteam/structurize/placement/handlers/entity/EntityHandlers.java
uecasm: Yeah, that makes no sense to me; collecting those items is kind of the whole point.

Nightenom: It's done by default method, that does handle that uses the itemhandler extract method

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
Nightenom: This is just random refactor, I'm kinda sure it's correct, just java and nested for loops, we want to switch to normal registry stuff here, and extend it with findHandlerFor(T value) which will return proper instance - to avoid having the search stuff everywhere

PR #758, opened 2025-06-29
Use correct placement handler for removal

uecasm:
Closes [MineColonies issue](https://github.com/ldtteam/minecolonies/pull/10976#issuecomment-3017207654)

- Use the removal handler for the block being removed, not the block replacing it.

## Testing

Review please (do not port)

~Disclaimer: this is not tested that it was even the problem or that it solves it.  But it seems plausible.~
Edit: I've tested this in 1.21 and it does appear to fix the issue.

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
+++ b/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
@@ -28,6 +28,7 @@ import net.minecraft.world.level.block.state.properties.BlockStateProperties;
 import net.minecraft.world.level.block.state.properties.DoubleBlockHalf;
 import net.minecraft.world.level.block.state.properties.DripstoneThickness;
 import net.minecraft.world.phys.AABB;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -98,6 +99,29 @@ public static void add(IPlacementHandler handler)
...
+    {
+        for (final IPlacementHandler placementHandler : handlers)
+        {
+            if (placementHandler.canHandle(world, worldPos, newState))
+            {
+                return placementHandler;
+            }
+        }
+
+        return null;
```
Nightenom, line 122: This should never happen, there is the default handler that takes everything. I know the old code had it, but even then it's should unreachable

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -98,6 +99,29 @@ public static void add(IPlacementHandler handler)
...
+     * Finds the appropriate {@link IPlacementHandler} for the given location.
+     * @param world     the world.
+     * @param worldPos  the world position.
+     * @param newState  the blockstate being placed or removed.
+     * @return          the appropriate handler, or null.
+     */
+    @Nullable
+    public static IPlacementHandler getHandler(@NotNull final Level world,
+                                               @NotNull final BlockPos worldPos,
+                                               @NotNull final BlockState newState)
```
Nightenom, line 112: I assume notnull by default, it's kinda noisy to have it everywhere

Nightenom (commented):
Approvable

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Raycoms: We could add a log here to catch this, yeah

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Raycoms: It's mostly our code style tbh

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
uecasm: You're correct, but the old code did admit it as a possibility.  I could change it to throw instead if it does happen.

Raycoms (approved):

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
Raycoms: Please no throw, just an error log

PR #764, opened 2025-07-28
Highlight box rendering

someaddons:
Remove background from text rendering
Allow boxes to render through blocks
image

## Testing

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
+++ b/src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
@@ -2,14 +2,8 @@ package com.ldtteam.structurize.util;
 
 import com.ldtteam.blockui.UiRenderMacros;
 import com.mojang.blaze3d.systems.RenderSystem;
-import com.mojang.blaze3d.vertex.BufferBuilder;
-import com.mojang.blaze3d.vertex.DefaultVertexFormat;
-import com.mojang.blaze3d.vertex.PoseStack;
-import com.mojang.blaze3d.vertex.Tesselator;
-import com.mojang.blaze3d.vertex.VertexConsumer;
-import com.mojang.blaze3d.vertex.VertexFormat;
-import org.joml.Matrix4f;
-import org.lwjgl.opengl.GL11;
+import com.mojang.blaze3d.vertex.*;
+import it.unimi.dsi.fastutil.objects.Object2ObjectLinkedOpenHashMap;
 import net.minecraft.Util;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.Font;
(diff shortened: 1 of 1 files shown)


src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
```diff
@@ -976,7 +972,7 @@ public static void renderDebugText(final BlockPos renderPos,
             matrixStack.mulPose(erm.cameraOrientation());
             matrixStack.scale(-0.014f, -0.014f, 0.014f);
 
-            final float backgroundTextOpacity = Minecraft.getInstance().options.getBackgroundOpacity(0.25F);
+            final float backgroundTextOpacity = 0f;
```
Nightenom, line 975: I'm kinda against this, one just don't render white text on light background

src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
```diff
@@ -1093,7 +1089,7 @@ private RenderTypes(final String nameIn,
                 .setTextureState(NO_TEXTURE)
                 .setShaderState(POSITION_COLOR_SHADER)
                 .setTransparencyState(TRANSLUCENT_TRANSPARENCY)
-                .setDepthTestState(LEQUAL_DEPTH_TEST)
+                .setDepthTestState(AlwaysDepthTestStateShard.ALWAYS_DEPTH_TEST)
```
Nightenom, line 1092: This might (will) require changes elsewhere, cuz sorting. I kinda don't care whether we render whole or not, but it might be kinda noisy if too many are on screen

src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
someaddons: the way minecraft renders the background behind the text, just messes up the text. Without it is always better

src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
someaddons: we'll see if it gets too noise with multiple schematics, though the tag tool rendering does require it to be able to see tagged blocks behind walls

src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
Nightenom: That's bug in forge, and there was a pr upstream to fix it, please do not port to 1.21.
Also without it it's 50% times worse, not sure how you can see white on bright colors, but I don't

src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
Nightenom: The single block highlights need adjusting, Mirality already had some decent ideas in mcol

src/main/java/com/ldtteam/structurize/util/WorldRenderMacros.java
someaddons: 1.21 had the exact same issue

PR #768, opened 2025-08-10
Tagoptions

Raycoms:
Closes #
Closes #

- Add Registered Tag Options to the tag tool UI
-

## Testing

Review please


src/main/java/com/ldtteam/structurize/client/gui/WindowTagTool.java
```diff
@@ -16,22 +17,33 @@
...
+import java.util.*;
+
+import static com.ldtteam.structurize.api.constants.Constants.GROUNDLEVEL_TAG;
 
 public class WindowTagTool extends AbstractWindowSkeleton
 {
+    /**
+     * List of tag options. Mods can just insert on this in mod constructor.
+     */
+    public static List<String> TAG_OPTIONS = new ArrayList<>();
```
Thodor12, line 29: If the intention is that other mods should be writing into this list, maybe for safety make it a concurrent list?

Thodor12: (Also probably want a method to add into it, to prevent mods from also clearing the list)

Raycoms: That should be fine. If any mod writes outside of the main thread or pretty much non static init into it. Sth is off.

Raycoms: They could just AT, mixin, whatever, if they wanted to anything more aggressive.

Thodor12: Fair enough

Thodor12 (dismissed):

src/main/java/com/ldtteam/structurize/client/gui/WindowTagTool.java
```diff
@@ -200,4 +226,31 @@ public void updateElement(final int index, final Pane rowPane)
...
+                return TAG_OPTIONS.size();
+            }
+
+            @Override
+            public void updateElement(final int index, final Pane rowPane)
+            {
+                final Text tagsText = rowPane.findPaneOfTypeByID(TAG_TEXT, Text.class);
+                tagsText.setText(Component.literal(TAG_OPTIONS.get(index)));
+                PaneBuilders.tooltipBuilder().hoverPane(tagsText).build()
+                    .setText(Component.translatable("com.ldtteam.tag.tooltip." + TAG_OPTIONS.get(index)));
```
Nightenom, line 249:
```suggestion
                    .setText(Component.translatable("structurize.gui.tagtool.tooltip." + TAG_OPTIONS.get(index)));
```

Raycoms: I was thinking about this, but I want to be able to add tooltips from minecolonies side etc as well and making it com.ldtteam.tag would feel cleaner then

src/main/resources/assets/structurize/gui/windowtagtool.xml
```diff
@@ -14,6 +14,15 @@
         </box>
     </list>
 
+    <text id="taglistname" size="130 12" pos="0 65" color="white" label="$(com.ldtteam.structurize.gui.tagoptions.title)"/>
+    <list id="tagoptionlist" size="130 120" pos="0 80" childspacing="-1">
+        <box size="100% 17" linewidth="1">
+            <text id="tagnames" size="190 25" pos="5 5" textalign="TOP_LEFT" color="white"/>
+            <button id="select" pos="104 1" size="16 16" source="structurize:textures/gui/builderhut/checkbox_empty.png" textcolor="black"
```
Nightenom, line 21: does this button have handler? if not change cursor to default

src/main/java/com/ldtteam/structurize/client/gui/WindowTagTool.java
Nightenom: well, i rather use modid in strings so it's always clean who is the owner

src/main/resources/assets/structurize/gui/windowtagtool.xml
Raycoms: yes it does

src/main/resources/assets/structurize/lang/en_us.json
```diff
@@ -229,6 +229,7 @@
   "structurize.gui.content.entities": "Entities",
   "structurize.gui.content.entities_content": "Entity contents",
   "com.ldtteam.structurize.gui.taglist.title": "Existing tags in schematic:",
-  "com.ldtteam.structurize.gui.tagoptions.title": "Tag Suggestions:"
+  "com.ldtteam.structurize.gui.tagoptions.title": "Tag Suggestions:",
+  "com.ldtteam.tag.tooltip.groundlevel": "Buildtool aligns the blueprint with the floor with the help of this tag"
```
Nightenom: I would rather go with sth: tagged block gets aligned with average ground level
"with the help of this tag" seems useless

src/main/java/com/ldtteam/structurize/client/gui/WindowTagTool.java
Raycoms: Yes, but here the owner is not structurize. It's the mod that adds the tag option. That mod will add it to their lang file too.

Nightenom: the owner is structurize, cuz without it there is not lang string :D

Nightenom (approved):

someaddons (approved):

src/main/java/com/ldtteam/structurize/client/gui/WindowTagTool.java
uecasm, line 29: I don't like this being a single static list.  The idea I was thinking about earlier was to have an interface method on the blueprint interface (i.e. anchor block where tags are being saved) so that the list could be building-specific.

Default interface implementation in Structurize could return groundlevel/invisible and other global tags.  Deco controller and hut blocks can provide their specific tags.

Raycoms: We could make this the global tag list and add an additional list with a "predicate, tag" tuple, where the predicate returns true if it applies to an anchor block.

uecasm: Why not just put conditional logic in AbstractHutBlock for the tags that are shared by many buildings?  Seems tidier than a flat list with predicates.

Raycoms: Because deco controllers also share most of those tags (sit/stand) for example

uecasm: They could both call a common utility method then .

Raycoms: I'd like to have most of the tag lists in the same place. So the option is either a static list in each of the individual classes we fill from the same location and query. Or we could have a predicate we fill at that place and just put it directly into structurize. I find the latter a bit more simple

PR #771, opened 2025-08-10
Tag manager

Raycoms:
Closes #
Closes #

-
-

## Testing

Review please


Thodor12 (dismissed):

src/main/java/com/ldtteam/structurize/client/gui/WindowTagTool.java
```diff
@@ -91,13 +85,23 @@ public WindowTagTool(String currentTag, BlockPos anchorPos, final Level world, f
         this.anchorPos = anchorPos;
         this.stack = stack;
 
+        tagOptions.addAll(TagManager.getGlobalTagOptions());
+
+        if (anchorPos != null)
+        {
+            final Block block = world.getBlockState(anchorPos).getBlock();
```
someaddons: you need the BE not the block

src/main/java/com/ldtteam/structurize/api/TagManager.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Register a block specific tag option.
+     * @param tag the tag to register.
+     * @param predicate the predicate that has to apply.
+     */
+    public static void registerSpecificTagOption(final String tag, final Predicate<IBlueprintDataProviderBE> predicate)
+    {
+        BLOCK_SPECIFIC_TAG_OPTIONS.put(tag, predicate);
```
uecasm: This should detect if the tag already exists in the map, and then join the existing and new predicates with an OR.

That would make it easier for addon mods to augment existing tags with their own building types.  (Though I still like the idea of making it an interface method better.)

Raycoms: Done

src/main/java/com/ldtteam/structurize/api/TagManager.java
```diff
@@ -0,0 +1,79 @@
...
+    public static void registerSpecificTagOption(final String tag, final Predicate<IAnchorBlock> predicate)
+    {
+        if (BLOCK_SPECIFIC_TAG_OPTIONS.containsKey(tag))
+        {
+            BLOCK_SPECIFIC_TAG_OPTIONS.put(tag, BLOCK_SPECIFIC_TAG_OPTIONS.get(tag).or(predicate));
+        }
+        else
+        {
+            BLOCK_SPECIFIC_TAG_OPTIONS.put(tag, predicate);
+        }
```
uecasm: `BLOCK_SPECIFIC_TAG_OPTIONS.compute(tag, (k, v) -> (v == null) ? predicate : v.or(predicate));`

someaddons: actually its nicer to read with more lines

Raycoms: I know I could one line it, but this way it's a bit more obvious what it is doing.

uecasm: `compute` is more efficient though.

Raycoms: This is a one time thing on startup

uecasm: Then at least put `getOrDefault(tag)` in a variable and check that for null instead of the contains check, to prevent double-dipping.

src/main/java/com/ldtteam/structurize/api/TagManager.java
```diff
@@ -0,0 +1,79 @@
...
+        GLOBAL_TAG_OPTIONS.add(GROUNDLEVEL_TAG);
+        GLOBAL_TAG_OPTIONS.add(INVISIBLE_TAG);
+    }
+    /**
+     * Register a new global tag option.
+     * @param tag the option to register.
+     */
+    public static void registerGlobalTagOption(final String tag)
+    {
+        if (!GLOBAL_TAG_OPTIONS.contains(tag))
```
Thodor12: If it's meant to be unique, might as well use a Set then

someaddons (dismissed):

PR #777, opened 2025-08-14
Fix allowPlayerSchematics

### LINKED ISSUE #776: [BUG] allowPlayerSchematics not working

- [x] I have searched the existing issues.

- [x] I am running the latest **beta/release** version of Structurize for my Minecraft version.

- [x] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/).

### Minecraft Version

1.20

### Structurize Version

1.20.1-1.0.780-snapshot

1.20.1-1.1.1010-snapshot

forge 47.4.0
blockui 1.20.1-1.0.193
domum_ornamentum 1.20.1-1.0.290-snapshot

### Current Behavior

I changed allowPlayerSchematics to false, but players can still choose schematics they created themselves

### Expected Behavior

Players cannot use schematics created by themselves

### Reproduction Steps

1. Install the mod on the server
2. Start the server, and after the configuration file is generated, shut down the server
3. Edit structurize-server.toml and set allowPlayerSchematics to false

### Logs

no logs

### Anything else?

_No response_

### Footer

---
## Viewers

Thodor12:
Closes #776 

- First, I noticed that locally scanned schematics still could be uploaded to the server despite the config being on, the option was only attempting to hide local schematics, not actually blocking them server side if they still managed to show up somehow.
- Second, due to an early return if, local only schematic packs weren't actually being hidden as they should have, allowing the config option to be bypassed. This early return has been removed so each pack is compared against the server to ensure they exist on either side at all times.
- Third, as per discussion with Sam we opted to turn the configuration option defaulting to **false** given the extra security risk of allowing people to upload arbitrary files directly to the server.
- Fourth, to combat concerns of packs being properly disabled and scanning not working, etc. We've opted to change how pack disabling works:
  - When a pack now gets disabled, you still see it in the list (compared to the prior plain removal), however they're flagged as disabled, so you cannot select the pack, stating why you cannot select the pack.
  - This also changes some code regarding random pack selection, since we can't select disabled packs.
  - Random pack selection can now yield no results, when there ARE packs available, just disabled. So the "no packs installed" message had to be removed. Instead you are now shown the pack selection screen directly when it cannot choose a random pack.
  - At the off-chance that the switch pack screen were to be fully empty, you'd see a text stating there's no packs.
    - After testing however, I found that the client always creates your user pack even if you didn't do anything, so the text realistically cannot display, might need to remove the automatic local pack creation if you're not scanning anything.
  
Other updates:
- The "cancel" button in the pack selection screen never functioned because an overlapping label was blocking click handlers from triggering the button
- The "selected pack" caching is now removed from the build tool window/StructurePacks class, due to the fact that these classes were getting overridden, they wouldn't be the same instance at all times, causing some weird behaviour. Selected pack has been turned into the name of the pack itself, given that the actual pack meta can be fetched with an O(1) call.
- Turned the client configuration option for "scan tool scrolling" off by default as per request by Ray.

## Testing
- [X] I also did a multiplayer test.

Review please


src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
```diff
@@ -179,18 +179,6 @@ else if (Minecraft.getInstance().level == null && (loadingState == ClientLoading
     public static void onServerSyncAttempt(final Map<String, Double> serverStructurePacks)
     {
         Network.getNetwork().sendToServer(new SyncSettingsToServer());
-
-        if (serverStructurePacks.isEmpty())
```
Raycoms, line 183: We should keep the early return and rather extract the disable logic into a method and call it on early return as well. Otherwise we run into sync issues.

Talyda:
Will the config being defaulted to "false" affect all deco scans as well? Or is it just for actual levelable builds?

uecasm:
This needs to be tested with the shape tool as well.  That currently works by generating the blueprint client-side and relying on the sync mechanism to let builders load it from the server.

Talyda:
That was actually what made me ask, cos I know we get a second client side folder with shape tool

src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
```diff
@@ -241,6 +221,38 @@ else if (version != pack.getVersion())
...
+        for (final StructurePackMeta pack : StructurePacks.getPackMetas())
+        {
+            final double version = serverStructurePacks.getOrDefault(pack.getName(), -1.0);
+            if (version == -1)
+            {
+                if (!Structurize.getConfig().getServer().allowPlayerSchematics.get())
+                {
+                    // Don't have this pack on the server, disable.
+                    StructurePacks.disablePack(pack.getName());
+                }
```
Raycoms, line 233: Did you test this with a local style jar? If it works as expected?

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
```diff
@@ -68,7 +68,7 @@ protected ServerConfiguration(final ForgeConfigSpec.Builder builder)
         createCategory(builder, "gameplay");
 
         ignoreSchematicsFromJar = defineBoolean(builder, "ignoreSchematicsFromJar", false);
-        allowPlayerSchematics = defineBoolean(builder, "allowPlayerSchematics", true);
+        allowPlayerSchematics = defineBoolean(builder, "allowPlayerSchematics", false);
```
uecasm: Yeah, changing this default is definitely wrong.

Thodor12: It's not wrong, check the PR description. The default is turned off as by nature there's a security risk of allowing the client to upload any random file to the server. They don't need to be uploading valid blueprints, it could be some sort of virus.

Having the default turned off is safer, and servers/communities that do want to use it, have the freedom to turn it back on (i.e. schematic servers or trusted communities).

Thodor12:
> Will the config being defaulted to "false" affect all deco scans as well? Or is it just for actual levelable builds?

> This needs to be tested with the shape tool as well. That currently works by generating the blueprint client-side and relying on the sync mechanism to let builders load it from the server.

Sam asked me this too in DMs and I confirmed that all scan tool scans are local only, so shape tool works the same way.
This also means that, if the config option is off, that one cannot use the scan/shape tool to create custom scans and use them.

I wasn't personally fully happy with this change, and I can see room for a separate config option to be introduced to still allow scans, but still don't allow client side packs. But for right now we deemed this change acceptable, as the whole point of this option is to not allow the player to bring any of their own schematics to the table, and only utilize what the server allows you to bring.

So the option now works as intended.

Raycoms:
> > Will the config being defaulted to "false" affect all deco scans as well? Or is it just for actual levelable builds?
> 
> > This needs to be tested with the shape tool as well. That currently works by generating the blueprint client-side and relying on the sync mechanism to let builders load it from the server.
> 
> Sam asked me this too in DMs and I confirmed that all scan tool scans are local only, so shape tool works the same way. This also means that, if the config option is off, that one cannot use the scan/shape tool to create custom scans and use them.
> 
> I wasn't personally fully happy with this change, and I can see room for a separate config option to be introduced to still allow scans, but still don't allow client side packs. But for right now we deemed this change acceptable, as the whole point of this option is to not allow the player to bring any of their own schematics to the table, and only utilize what the server allows you to bring.
> 
> So the option now works as intended.

To add to this, is that especially on public servers there are some hard to control, esp in big pack bypass ways to cheat things in. So by default it's off and a server-owner with smaller private groups can just turn it on.

MotionlessTrain:
I’m starting to mentally prepare for all questions why the scan tool and shape tool are not working on a newly set up server
Maybe the crafting recipes of those items need to be disabled if the config is off, if able? As the items don’t really work anyway (as the personal style pack is a local style pack, so it may not be used)

Thodor12:
> I’m starting to mentally prepare for all questions why the scan tool and shape tool are not working on a newly set up server Maybe the crafting recipes of those items need to be disabled if the config is off, if able? As the items don’t really work anyway (as the personal style pack is a local style pack, so it may not be used)

At the very least I can confirm that local packs are **not** visible in the build tool. So if you were to scan something in locally, you wouldn't see your personal pack in the list.

Raycoms:
> I’m starting to mentally prepare for all questions why the scan tool and shape tool are not working on a newly set up server Maybe the crafting recipes of those items need to be disabled if the config is off, if able? As the items don’t really work anyway (as the personal style pack is a local style pack, so it may not be used)

Thanks for mentioning that. That reminds me. @Thodor12 can we add a msg to the player if they try to scan or buildtool place. Maybe we don't disable the player pack after all, but do a "msg on attempt" to the player with "enable config on server".

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
uecasm: They're data files, it's literally impossible for it to be a virus.  The only possible route of attack is via a buffer overrun smashing the stack and producing executable code, which is incredibly hard to achieve in the JVM.  And would be a bug if Structurize were vulnerable to that in the first place, which we should fix instead.

This is just going to annoy people who forget to manually turn it back on, for not really any benefit whatsoever; it's only useful for strongly themed servers that don't want any kind of custom decorations -- they're already protected against custom buildings regardless of this setting.

Raycoms: Yeah, it's not for virus. It's for flooding.

Raycoms: you can just flood a server with schematics

uecasm:
> Will the config being defaulted to "false" affect all deco scans as well? Or is it just for actual levelable builds?

Non-deco scans are blocked regardless of this setting.  If this setting is set false, it will additionally block player decorations *and* the shape tool.  Hence why I'm saying it should absolutely not be defaulted to false.

uecasm:
Also, if this is set to false, does it block the player pack in a single-player world as well?  Or does it only affect true servers?

Because I don't think it should block the local pack in SP regardless of this setting ... or if it does, another reason why it should not default to false.

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
someaddons: and it also protects against abuses like duping with custom schematics, which we had over the years multiple times

Raycoms:
> Also, if this is set to false, does it block the player pack in a single-player world as well? Or does it only affect true servers?
> 
> Because I don't think it should block the local pack in SP regardless of this setting ... or if it does, another reason why it should not default to false.

It should not affect singleplayer worlds @Thodor12 please make sure when testing this, that this works.

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
uecasm: Flooding can be solved by implementing a max blueprint count + filesize and deleting old ones.

Duping isn't an issue in Structurize, and for Minecolonies can be solved with placement handlers as discovered.

Thodor12:
Singleplayer doesn't get affected by this because of the shared filesystem, local packs are always server packs

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
someaddons: We solved it multiple times already, the feature simple has an inherent risk to it and as such should default to off as those which enable it should be aware of those risks

Talyda:
> I wasn't personally fully happy with this change, and I can see room for a separate config option to be introduced to still allow scans, but still don't allow client side packs.

What do you mean by allow scans, but not client side packs?  Client side packs can't be used anyway, they have to be on the server - only deco/scans can be used from client side.

I think separate option for the shape tool at the very least should be a thing - the shape tool is different to scans in a way, and is too useful to be totally blocked


> I’m starting to mentally prepare for all questions why the scan tool and shape tool are not working on a newly set up server

Yeah, me too, is why I am asking for clarification on this as it goes!

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
Raycoms: Especially with random mods there will always be a way to sneak things in. It's very difficult to properly handle this for all edge cases. So it's better to have this default off.

uecasm: sadface

Thodor12:
> > I wasn't personally fully happy with this change, and I can see room for a separate config option to be introduced to still allow scans, but still don't allow client side packs.
> 
> What do you mean by allow scans, but not client side packs? Client side packs can't be used anyway, they have to be on the server - only deco/scans can be used from client side.
> 
> I think separate option for the shape tool at the very least should be a thing - the shape tool is different to scans in a way, and is too useful to be totally blocked
> 
> > I’m starting to mentally prepare for all questions why the scan tool and shape tool are not working on a newly set up server
> 
> Yeah, me too, is why I am asking for clarification on this as it goes!

Scanning is something we can do directly from in-game, so we know what's going to happen, what we have, etc. This could be uploaded with less of a hassle compared to something that was created by someone outside of the game.

Raycoms:
The most important thing is to have some ingame way that tells players why it's not working. So that we don't get a lot of reports.

Raycoms:
Btw, can you also make scan tool scrolling default to off?

Raycoms:
> > > I wasn't personally fully happy with this change, and I can see room for a separate config option to be introduced to still allow scans, but still don't allow client side packs.
> > 
> > 
> > What do you mean by allow scans, but not client side packs? Client side packs can't be used anyway, they have to be on the server - only deco/scans can be used from client side.
> > I think separate option for the shape tool at the very least should be a thing - the shape tool is different to scans in a way, and is too useful to be totally blocked
> > > I’m starting to mentally prepare for all questions why the scan tool and shape tool are not working on a newly set up server
> > 
> > 
> > Yeah, me too, is why I am asking for clarification on this as it goes!
> 
> Scanning is something we can do directly from in-game, so we know what's going to happen, what we have, etc. This could be uploaded with less of a hassle compared to something that was created by someone outside of the game.

Might find a way to duplicate ingame items still though.

src/main/java/com/ldtteam/structurize/config/ServerConfiguration.java
ddaodan: > and it also protects against abuses like duping with custom schematics, which we had over the years multiple times

Yes, this is exactly why I urgently needed to disable it. I've already discovered a duplication bug

Thodor12:
@Talyda @MotionlessTrain Discussed with Ray, we're making it so the packs are still visible in the build tool, but disabled, and the selection buttons will show a warning on hover as to why they are disabled like so:

2025-08-15 15_01_14-Window

This should hopefully reduce some questions.

MotionlessTrain:
That would help, indeed

Talyda:
Oh, yep! That will be much better! Thanks!

Raycoms (dismissed):

someaddons (dismissed):

PR #783, opened 2025-09-22
Fix incorrect pack disabling

Thodor12:
Closes [discord report](https://discord.com/channels/139070364159311872/1125094832143077456/1419745392316121199) (all minecolonies pack are getting disabled)

- The synchronisations were only doing partial synchronisations of things it thought it had to sync, this was causing some packs to be incorrectly marked as disabled
- Synchronisation now syncs the full pack list in all cases, leaving the full comparison up to the server and client
- The server no longer determines to only sync missing packs back to the client, instead it sends everything, if something is missing, it'll still send the server->client pack transfer message out
- Fixed an issue where local mutable incorrectly versioned packs wouldn't get removed, and thus wouldn't be overwritten with a new version upon sync, these are now correctly removed instead of getting disabled (only immutable packs will be disabled, i.e. if a client brings a different version of Minecolonies somehow, and has a pack with a different verrsion)

## Testing
- [X] I also did a multiplayer test.

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
+++ b/src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
@@ -175,8 +175,6 @@ public class ClientStructurePackLoader
 
         if (serverStructurePacks.isEmpty())
         {
-            checkPackDifferences(serverStructurePacks);
-
             // Most likely single player. Skip.
             loadingState = ClientLoadingState.FINISHED_SYNCING;
             StructurePacks.setFinishedLoading();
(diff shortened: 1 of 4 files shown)


src/main/java/com/ldtteam/structurize/storage/ClientStructurePackLoader.java
```diff
@@ -218,22 +216,33 @@ private static boolean checkPackDifferences(final Map<String, Double> serverStru
...
+                if (pack.isImmutable())
+                {
+                    pack.setDisabled(true);
+                }
+                // If the pack is mutable, we remove the local copy and tell the server we want to synchronize this pack anew
+                // On the next iteration, the client should tell the server it's now missing this pack, which should initiate a copy from the server
+                else
                 {
+                    StructurePacks.removePack(pack.getName());
                     needsChanges = true;
```
uecasm, line 246: Doesn't this break the logic that `onStructurePackTransfer` uses to delete the old version of the pack?

Thodor12: Technically no, since the previous "StructurePacks.disablePack" was a removal

PR #815, opened 2026-01-27
Using ChunkStatus.SURFACE insteadof null

gisellevonbingen:
Closes #
Closes #

- RelativityMC/C2ME-fabric#508
-

## Testing

Crashed code.
https://github.com/RelativityMC/C2ME-fabric/blob/4e4b37e5848f70a4b53776d59263f831c6a7d113/c2me-threading-worldgen/src/main/java/com/ishland/c2me/threading/worldgen/mixin/MixinChunkRegion.java#L41-L44
image

The parameter what type is `ChunkStatus` is not marked as nullable.
I think using `ChunkStatus.EMPTY` is better than `null` even if it is no used anywhere in vanilla.
image

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
@@ -739,7 +739,7 @@ public final class BlockUtils
     {
         private OurWorldGenRegion(ServerLevel p_143484_, List<ChunkAccess> p_143485_)
         {
-            super(p_143484_, p_143485_, null, -1);
+            super(p_143484_, p_143485_, ChunkStatus.EMPTY, -1);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -739,7 +739,7 @@ private static class OurWorldGenRegion extends WorldGenRegion
     {
         private OurWorldGenRegion(ServerLevel p_143484_, List<ChunkAccess> p_143485_)
         {
-            super(p_143484_, p_143485_, null, -1);
+            super(p_143484_, p_143485_, ChunkStatus.EMPTY, -1);
```
Nightenom: Use SURFACE please, it's not used by vanilla anyway in 1.20

gisellevonbingen: Changed.

Raycoms (approved):

PR #818, opened 2026-02-08
rework world equals checks

Raycoms:
Closes #
Closes #

- Properly Handle Equal checks in the IPlacementHandlers
- Pass context in new "IPlacementContext"
- Fix solid/fluid/grass and potted plant oddities

This will break binary compat intentionally as otherwise it can cause weird bugs.

## Testing

Still testing don't merge yet

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
+++ b/src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
@@ -8,6 +8,7 @@ import com.ldtteam.structurize.blocks.ModBlocks;
 import com.ldtteam.structurize.placement.handlers.placement.IPlacementHandler;
 import com.ldtteam.structurize.placement.handlers.placement.PlacementHandlers;
 import com.ldtteam.structurize.placement.structure.IStructureHandler;
+import com.ldtteam.structurize.util.BlockInfo;
 import com.ldtteam.structurize.util.BlockUtils;
 import com.ldtteam.structurize.util.ChangeStorage;
 import net.minecraft.core.BlockPos;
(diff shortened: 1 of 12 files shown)


Thodor12:
@Raycoms What is this sonarqubecloud thing, is this something new?

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -471,7 +470,7 @@ public static boolean doBlocksMatch(final ItemStack block, final ServerLevel wor
         {
             final IPlacementHandler handler = PlacementHandlers.getHandler(world, BlockPos.ZERO, blockState);
             final List<ItemStack> itemList =
-              handler.getRequiredItems(world, position, blockState, tileEntity == null ? null : tileEntity.saveWithFullMetadata(world.registryAccess()), true);
+              handler.getRequiredItems(world, position, blockState, tileEntity == null ? null : tileEntity.saveWithFullMetadata(world.registryAccess()), null);
```
someaddons: null instead of true/false?

src/main/java/com/ldtteam/structurize/placement/StructurePlacer.java
someaddons, line 271: since we have like 3 handler lookups in a row, maybe we should cachte the result? Or add a Block -> handler hashmap cache to the Placementhandlers.getHandler(), which is cleared when adding a new handler

MotionlessTrain: I don’t supposed placementhandlers are added later, after blocks are being searched. I at least assumed that is a thing that would need to be done whilst Minecraft is starting

someaddons: in theory we could add a custom handler later from e.g. minecolonies so to avoid future potential issues that should clear the cache

Raycoms: The first handler we need to query, but I agree that the second one we should cache

someaddons (dismissed):

someaddons:
approved when the null thing is fixed

someaddons (approved):

PR #820, opened 2026-02-24
Synchronize access

Raycoms:
Closes #
Closes #

- Synchronize access to avoid concurrency bugs
-

## Testing

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
+++ b/src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
@@ -30,6 +30,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
+import java.util.concurrent.locks.Lock;
 
 import static com.ldtteam.structurize.api.constants.Constants.UPDATE_FLAG;
 
@@ -43,6 +44,7 @@ import static com.ldtteam.structurize.api.constants.Constants.UPDATE_FLAG;
 public final class PlacementHandlers
 {
     public static final List<IPlacementHandler> handlers = new ArrayList<>();
+
     static
     {
         handlers.add(new AirPlacementHandler());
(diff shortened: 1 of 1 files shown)


someaddons (approved):

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -97,8 +139,11 @@ public static void add(IPlacementHandler handler, Class<?> override)
      */
     public static void add(IPlacementHandler handler)
     {
-        handlers.add(1, handler);
-        handlerCache.clear();
+        synchronized (handlers)
+        {
+            handlers.add(1, handler);
```
uecasm, line 144: I realise you didn't change this, but why is this 1 rather than 0?

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
```diff
@@ -78,15 +84,51 @@ public final class PlacementHandlers
...
+                if (override == handlers.get(i).getClass())
+                {
+                    switch (addType)
+                    {
+                        case BEFORE:
+                            handlers.add(i - 1, handler);
+                            break;
+                        case AFTER:
+                            handlers.add(i, handler);
+                            break;
```
uecasm, line 122: This is incorrect; BEFORE should be i and AFTER should be i + 1.

src/main/java/com/ldtteam/structurize/placement/handlers/placement/PlacementHandlers.java
MotionlessTrain: I guess that the AirPlacementHandler should be able to be overridden by adding a new placement handler

Raycoms: It was for performance reasons, most common ones first.

PR #825, opened 2026-03-20
Merge Highlightmanager into structurize

someaddons:
Closes #
Closes #

- Merges minecolonies rendering util highlightmanager into structurize, as both mods are doing similar things(box/block rendering)
- Migrated WorldEventRenderContext to structurize
- Merged and moved rendertypes into its own file
- Added unified RenderTaskManager which handles all render tasks, draw calls and ticking
- Add default tasks for timed tasks, Box tasks and blockpos render tasks with text.

## Testing

Review please


Raycoms (commented):
Is this compat with Minecolonies or will this require changes there too?

someaddons:
hm gotta test, depends on what minecolonies used as I moved 2-3 classes around

PR #828, opened 2026-04-21
Allow support blocks on tan placeholder faces

uecasm:
Closes [discord request](https://discord.com/channels/472875599422291968/938589424441233458/1496101864100466748)

- Allows supported blocks to be placed on tan placeholders, without changing their normal collision behaviour.
    - Might not work in all cases due to mojank though.
image
- Also preserves farmland and paths underneath.
image

## Testing

Review please (should port)

### DIFF
--- a/src/main/java/com/ldtteam/structurize/blocks/schematic/BlockSubstitution.java
+++ b/src/main/java/com/ldtteam/structurize/blocks/schematic/BlockSubstitution.java
@@ -58,4 +58,11 @@ public class BlockSubstitution extends Block
 
         return super.getCollisionShape(state,blockGetter,blockPos,context);
     }
+
+    @Override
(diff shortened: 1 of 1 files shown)


Raycoms (dismissed):

someaddons (approved):

Shadizar:
won't this have a  high risk of things popping off all the time from things that can't support them?

uecasm:
> won't this have a high risk of things popping off all the time from things that can't support them?

The point is to put them in your L2+ so that you know there's something good for it to attach to (but doesn't count as a "solid" block so you can't use that placeholder instead).

But yes, if misused, it will pop blocks.  Don't do that.

Raycoms:
Port please

PR #832, opened 2026-05-16
Fix for possibly Arabic digits and other weird player filename combos.

uecasm:
Closes [discord report](https://discord.com/channels/139070364159311872/1125094832143077456/1505041198472233160)

- Fix one similar case to ldtteam/minecolonies#11651
- Fix player names not being escaped when turning into directory names

## Testing

Review please (should port)

### DIFF
--- a/src/api/java/com/ldtteam/structurize/api/util/Utils.java
+++ b/src/api/java/com/ldtteam/structurize/api/util/Utils.java
@@ -81,8 +81,8 @@ public final class Utils
      * @param input a player name or other possibly unsafe text.
      * @return the safe filename.
      *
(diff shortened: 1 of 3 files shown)


Raycoms (dismissed):

someaddons (approved):
Make sure we wrap all calls that use the name for the folder, else some may break

uecasm:
> Make sure we wrap all calls that use the name for the folder, else some may break

I did check everywhere that uses `Minecraft.getInstance().getUser().getName()`, and a more quick check for anything similar.  As far as I know, I got everywhere.  Let me know if I missed something.

Raycoms:
please port

PR #839, opened 2026-05-23
Fix block placement handler

### LINKED ISSUE #812: Scan tool replaced the "base" panel variant, rather than the variant selected when replacing trap doors

- [x] I have searched the existing issues.

- [x] I am running the latest **beta/release** version of Structurize for my Minecraft version.

- [x] I checked the MineColonies/Structurize [Wiki](https://wiki.minecolonies.ldtteam.com/) or I asked on [discord](https://discord.minecolonies.com/).

### Minecraft Version

1.20

### Structurize Version

structurize-1.20.1-1.0.784-snapshot

minecolonies-1.20.1-1.1.1065-snapshot

(Minecolonies Schematic server pack)

### Current Behavior

When replaced dark oak trap doors with horizontally stripped dark oak plank panels, it replaced all of them with "full" panels instead of the stripped.

Image

Image

Image

### Expected Behavior

Properly replaced with the variant selected.

### Reproduction Steps

Make a line of dark oak trap doors, try replacing them with horizontal panels with the scan tools.

### Logs

none

### Anything else?

Moobien asked me to make the report.

### Footer

---
## Viewers

Thodor12:
Closes #812

- Use the BlockItem#place method similar logic instead of custom placement logic

## Testing

Review please

### DIFF
--- a/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
+++ b/src/main/java/com/ldtteam/structurize/util/BlockUtils.java
@@ -573,35 +573,17 @@ public final class BlockUtils
         {
             world.removeBlock(here, false);
         }
-        else if (item instanceof BlockItem)
+        else if (item instanceof BlockItem blockItem)
         {
-            final Block targetBlock = ((BlockItem) item).getBlock();
-            BlockState newState = copyFirstCommonBlockStateProperties(targetBlock.defaultBlockState(), blockState);
-
-            if (newState == null)
-            {
-                fakePlayer.setItemInHand(InteractionHand.MAIN_HAND, stackToPlace);
-                if (stackToPlace.is(ItemTags.BEDS) && blockState.hasProperty(HorizontalDirectionalBlock.FACING))
-                {
-                    fakePlayer.setYRot(blockState.getValue(HorizontalDirectionalBlock.FACING).get2DDataValue() * 90);
-                }
-
-                newState = targetBlock.getStateForPlacement(new BlockPlaceContext(new UseOnContext(fakePlayer,
-                    InteractionHand.MAIN_HAND,
-                    new BlockHitResult(new Vec3(0, 0, 0),
-                        itemStack.getItem() instanceof BedItem ? Direction.UP : Direction.NORTH,
-                        here,
-                        true))));
-
-                if (newState == null)
-                {
-                    return;
-                }
-            }
-
-            // place
-            world.setBlock(here, Blocks.COBBLESTONE.defaultBlockState(), Block.UPDATE_CLIENTS);
-            world.setBlock(here, newState, Constants.UPDATE_FLAG);
+            final Block targetBlock = blockItem.getBlock();
+            fakePlayer.setItemInHand(InteractionHand.MAIN_HAND, stackToPlace);
+            final BlockPlaceContext ctx = new BlockPlaceContext(new UseOnContext(fakePlayer, InteractionHand.MAIN_HAND,
+                new BlockHitResult(new Vec3(0, 0, 0), Direction.NORTH, here, true)));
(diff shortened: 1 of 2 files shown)


MotionlessTrain:
Could you double check whether anything changed regarding torches which were standing on the block?
Just in case they first managed to stand, but now pop off

Thodor12:
This is only used during block replacement

MotionlessTrain:
Yeah, that is what I am talking about
When using this to replace a wall of your building with a lower tier material (when designing a lower level building), it would be annoying to pick up all torches that fell inside, and put them back at the right place
Iirc Structurize has support for that (no pun intended), and I was wondering whether that was this code

Thodor12:
Oh yeah that'll still be broken but that's impossible to fix because we have to give a block update unfortunately

MotionlessTrain:
As long as that doesn’t get worse than it is now

src/main/java/com/ldtteam/structurize/util/BlockUtils.java
```diff
@@ -573,35 +573,17 @@ public static void handleCorrectBlockPlacement(
...
-            }
-
-            // place
-            world.setBlock(here, Blocks.COBBLESTONE.defaultBlockState(), Block.UPDATE_CLIENTS);
-            world.setBlock(here, newState, Constants.UPDATE_FLAG);
+        else if (item instanceof BlockItem blockItem)
+        {
+            final Block targetBlock = blockItem.getBlock();
+            fakePlayer.setItemInHand(InteractionHand.MAIN_HAND, stackToPlace);
+            final BlockPlaceContext ctx = new BlockPlaceContext(new UseOnContext(fakePlayer, InteractionHand.MAIN_HAND,
```
Raycoms, line 580: We shouldn't go over this placement context because this will mess with blocks like stairs, etc as they use hit directions etc which will result in strange things.
Which of these things here specifically do the necessary things for the panels to work?

Thodor12: The problem is I need getStateForPlacement, which is what requires the placement context

Thodor12: I can do a test with stairs but important is that this is only used by scan tool replaces it looks like

Thodor12: And the copy properties method that goes after the placement state should copy the direction of a stair over

Raycoms: getStateForPlacement probably translates some meta data into the blockstate property we need. We should be able to get this from the internal call that method uses

MotionlessTrain: Just in case, according to documentation (that is, the chat message in game), if you replace a block with another block with similar block states, it should inherit the block states
If the new block has block states that are not present on the old block, the default block state property is used (and the chat message is printed)
Which means, in particular for stairs, if you are replacing one stair with another, it should have the same shape in the same direction, and if you are replacing a non-stair non-shingle with a stair, it should always face in the default direction (north?), independent from how you are standing

Thodor12: > getStateForPlacement probably translates some meta data into the blockstate property we need. We should be able to get this from the internal call that method uses

There's nothing else for this, check PanelBlock in DO

Thodor12: I verified and stairs continue to work normal. Because of:
```java
BlockState newState = copyFirstCommonBlockStateProperties(placementState != null ? placementState : targetBlock.defaultBlockState(), blockState);
```

